Issue #13, opened 2022-03-03
[Mostly solved; postponed] Make the code with type-level magic more readable and simpler


Mikolaj:
Here are the samples from the library code that only library maintainers are supposed to write:

https://github.com/Mikolaj/horde-ad/blob/93f1d067f7634c6505bcf4f2dc8614c8a302a5f8/src/HordeAd/Core/Delta.hs#L452-L454

https://github.com/Mikolaj/horde-ad/blob/93f1d067f7634c6505bcf4f2dc8614c8a302a5f8/src/HordeAd/Core/Delta.hs#L464-L479

And here is the code that, potentially, library users may want to write, extend, override:

https://github.com/Mikolaj/horde-ad/blob/93f1d067f7634c6505bcf4f2dc8614c8a302a5f8/src/HordeAd/Core/DualNumber.hs#L4-L5

https://github.com/Mikolaj/horde-ad/blob/93f1d067f7634c6505bcf4f2dc8614c8a302a5f8/src/HordeAd/Core/DualNumber.hs#L306-L316

Could any of this be simplified? Can the GHC plugins be avoided, could this be ported to GHC 8.10, can the auxiliary functions be avoided and the code written inline in the case expression instead?

Here's the library and the types underneath: https://hackage.haskell.org/package/orthotope-0.1.2.0/docs/Data-Array-Shaped.html

Mikolaj:
Related design questions below (to be answered via experiments, but these can be informed in advance by experience with type level programming in other projects).

* Currently we have a strange mix: totally untyped DeltaX, de-facto rank-typed Delta0, Delta1 and Delta 2, fully typed (shape == rank and size) DeltaS. However, debugging sizes turns out to be a pain, so we need shapes also for matrices and vectors. So perhaps use DeltaS instead of matrices? Or should we rather make matrices shaped? 
* Or perhaps place all objective function parameters (including matrices and vectors) in the vector of DeltaS tensors, move them around as shaped and convert them to matrices just before matrix multiplication and quickly convert, e.g., the resulting vector again into a shaped tensor?
* If we stick to only rank-typed matrices and vectors, should we replace Delta0 (unlikely, would cause 100 times slowdown), Delta1, Delta2 with (Delta (n::Nat)), e.g., using the rank-typed tensors from orthotope and so adding a third kind of sensors into the mix? Or remove the fully untyped DeltaX at that point?

goldfirere:
[...]

Issue #14, opened 2022-03-03
SPECIALIZE needed despite no warnings by -Wall-missed-specialisations


Mikolaj:
In short: despite `-fexpose-all-unfoldings -fspecialise-aggressively`, two `SPECIALIZE` pragmas are required to significantly speed up and lower allocation for the code.  Confirmed with GHC 8.10.7, 9.0.2 and 9.2.1.

That would imply that there is a specialization that is not performed by GHC without the pragmas. However, despite `-Wall-missed-specialisations`, no missed specialization is reported in the code without `SPECIALIZE` (with 8.10.7 and 9.0.2; ~OTOH 9.2.1 spams so much that it's hard to tell; we should try 9.2.2 once it's out~ [edit: 9.2.2 repro done, see below; in short: less spam, but no relevant warning]).

It would be great to understand if these are GHC bugs that needs to be reported and whether we may have other not applied specilizations that `-Wall-missed-specialisations` is not telling us about. Is there a cheap way to find them? Can we fix them or rule them out somehow even without finding them individually? Any there any ideas emerging from this case about how to speed up this particular part of code (which is computing the value of the objective function, ignoring derivatives; #11)?

Reproducing **BENCH1** with ghc-9.0.2:

```
git clone git@github.com:Mikolaj/horde-ad.git
git checkout ghc-report-abstract
cabal build --enable-benchmarks
`cabal list-bin mnist` -n 10 -m prefix "2-hidden-layer MNIST nn with samples: 500/test 500" +RTS -s
```

This provides the baseline, in which things are apparently not specialized. Note that there is no warning about missing specialization. For me, the results are

```
 509,766,256,688 bytes allocated in the heap
  Total   time  151.546s  (151.631s elapsed)
```

The following only adds a commit that introduces the two SPECIALIZE pragmas. [Edit: and then two others that I added only to eliminate thunks that would obscure what's going on, No significant change of results due to their elimination.]

```
git checkout ghc-report-specialize
cabal build --enable-benchmarks
`cabal list-bin mnist` -n 10 -m prefix "2-hidden-layer MNIST nn with samples: 500/test 500" +RTS -s
```

My results are

```
   2,463,534,528 bytes allocated in the heap
  Total   time    8.750s  (  8.757s elapsed)
```

[...]

PR #16, opened 2022-03-09
Don't use type variables in patterns

goldfirere:
This PR is an attempt to fix part #13 (getting GHC 8.10.7 compatibility). But it doesn't work, because of an overlapping instance. The sad thing is that I think there really is an overlapping instance. I will report a GHC bug about this shortly.

### DIFF
--- a/horde-ad.cabal
+++ b/horde-ad.cabal
@@ -132,7 +132,7 @@ library
     -- Other library packages from which modules are imported.
     build-depends:
         assert-failure
-      , base >= 4.16 && < 99
+      , base >= 4.14 && < 99
       , bytestring
       , ghc-typelits-knownnat
       , ghc-typelits-natnormalise
--- a/src/HordeAd/Core/Delta.hs
+++ b/src/HordeAd/Core/Delta.hs
@@ -49,6 +49,7 @@ import qualified Data.Array.Internal.DynamicS
 import qualified Data.Array.ShapedS as OS
 import           Data.Kind (Type)
 import           Data.List (foldl')
+import           Data.Proxy (Proxy(Proxy))
 import qualified Data.Strict.Vector as Data.Vector
 import qualified Data.Strict.Vector.Autogen.Mutable as Data.Vector.Mutable
 import qualified Data.Vector.Generic as V
@@ -182,13 +183,13 @@ data DeltaS :: [Nat] -> Type -> Type where
     -- ^ Append two arrays along the outermost dimension.
   SliceS :: forall i n k rest r.
             (KnownNat i, KnownNat n, KnownNat k, OS.Shape rest)
-         => DeltaS (i + n + k ': rest) r -> DeltaS (n ': rest) r
+         => Proxy i -> Proxy n -> DeltaS (i + n + k ': rest) r -> DeltaS (n ': rest) r
     -- ^ Extract a slice of an array along the outermost dimension.
 
   From0S :: Delta0 r -> DeltaS '[] r
   From1S :: Delta1 r -> DeltaS '[n] r
   From2S :: forall rows cols r. KnownNat cols
-         => Delta2 r -> DeltaS '[rows, cols] r
+         => Proxy cols -> Delta2 r -> DeltaS '[rows, cols] r
   FromXS :: DeltaX r -> DeltaS sh r
 
 instance Show (DeltaS sh r) where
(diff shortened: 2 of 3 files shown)


goldfirere:
The GHC bug is [reported upstream](https://gitlab.haskell.org/ghc/ghc/-/issues/21208). But we should still arrange not to have overlapping instances here.

Mikolaj:
Thanks a lot. Apologies for the conflict. Could you rebase and resolve the conflict so that CI tries out 8.10.7 and 9.0.2? Thank you.

Mikolaj:
Thank you. Now CI shows the overlap: https://github.com/Mikolaj/horde-ad/runs/5516892917?check_suite_focus=true#step:18:29

I don't understand much from this error message, though.

Mikolaj:
Might this be related to `AllowAmbiguousTypes` that was IIRC needed to define this `SliceS` constructor?

Mikolaj:
@goldfirere: while I'm clueless about the overlapping instances (all I know the class is defined here: https://hackage.haskell.org/package/orthotope-0.1.2.0/docs/Data-Array-Internal-Shape.html#t:Shape), I think the actual errors in 9.2.2 are due to a missing argument to `From2S` and so probably unrelated: https://github.com/Mikolaj/horde-ad/runs/5516892563?check_suite_focus=true#step:18:37

Do I get it right. Would it compile fine in GHC 9.2.2 once fixed? (And would the older GHCs compile fine with `IncoherentInstances`?)

How could we get rid of the overlapping instances?

Mikolaj:
Another, probably related, problem with SliceS: https://github.com/Mikolaj/horde-ad/blob/01f5b467d1b0ccf9e8e4244e0837829f857a6807/src/HordeAd/Core/HasDual.hs#L191

Uncommenting the type signature doesn't help:

```
src/HordeAd/Core/HasDual.hs:188:14: error:
    • Couldn't match type: (i0 + n) + k0
                     with: (i + n) + k
      Expected: Dual (OS.Array (((i + n) + k) : rest) Double)
                -> Dual (OS.Array (n : rest) Double)
        Actual: Dual (OS.Array (((i0 + n) + k0) : rest) Double)
                -> Dual (OS.Array (n : rest) Double)
      NB: ‘+’ is a non-injective type family
      The type variables ‘i0’, ‘k0’ are ambiguous
    • When checking that instance signature for ‘dSliceS’
        is more general than its signature in the class
        Instance sig: forall (i :: GHC.TypeNats.Nat)
                             (n :: GHC.TypeNats.Nat) (k :: GHC.TypeNats.Nat)
                             (rest :: [GHC.TypeNats.Nat]).
                      (KnownNat i, KnownNat n, KnownNat k, OS.Shape rest) =>
                      Dual (OS.Array (((i + n) + k) : rest) Double)
                      -> Dual (OS.Array (n : rest) Double)
           Class sig: forall (i :: GHC.TypeNats.Nat)
                             (n :: GHC.TypeNats.Nat) (k :: GHC.TypeNats.Nat)
                             (rest :: [GHC.TypeNats.Nat]).
                      (KnownNat i, KnownNat n, KnownNat k, OS.Shape rest) =>
                      Dual (OS.Array (((i + n) + k) : rest) Double)
                      -> Dual (OS.Array (n : rest) Double)
      In the instance declaration for ‘HasRanks Double’
    |
188 |   dSliceS :: forall i n k rest.
```

Mikolaj:
The `dSliceS` error from above is now reproducible on master branch (apologies for the merge conflicts again --- everything is now overhauled according to your ideas from last meeting --- the main class takes the dual component type as argument (e.g., Delta0 Double), not the primal (e.g., Double));

https://github.com/Mikolaj/horde-ad/blob/d069a45773ed849913b5ebd0345153072f304fd9/src/HordeAd/Core/DualClass.hs#L234-L238

Even the `dSliceS` signature alone suffices to cause an error. I wonder if the problem is going to vanish with the 9.0.2-compatible reformulation.

Mikolaj:
@goldfirere: I'm now blocked by this, because performance with GHC 9.2.2 is degraded and requires random hacks and with HEAD even more so (https://github.com/Mikolaj/horde-ad/issues/14), which makes performance measurements with GHC >= 9.2 suspect. Consequently, whenever I change anything in the code, I can't confidently benchmark the effects not having access to 9.0.2 at least. And I experienced with this codebase it's easy to gradually lose performance from little changes, feature additions or from optimizations that magnify unrelated pre-existing performance problems. If you are busy, I'd rebase and merge and try to complete this myself. I'm going to yell for help again, if I fail.

Mikolaj:
The rebase is going to be hopeless, so let me copy over the relevant bits and try to make it work.

Mikolaj:
I have manually rebased and completed this PR's work: d68277fb6b94907dfa8a7c5bebf0902f181f42a9 d68277fb6b94907dfa8a7c5bebf0902f181f42a9 cf6eaf7d392289bb7ebb14262a9c2e4439ee71b7

@goldfirere: That's so close to what you've proposed, including the IncoherentInstances under CPP that I'm not even asking for your review before closing. You may still have a chance to look at this code if you attempt to eliminate the auxiliary function that are needed only to capture proper types (I think the current proxies are not enough to fix that, but perhaps more proxies would be?) or to generally overhaul this type-level code if it's somehow fundamentally flawed.

Closing. Thank you!

Issue #29, opened 2022-04-28
Rename scalar to constant


tomjaguarpaw:
Shouldn't `scalar` be called `constant`? I doesn't seem to be restricted to scalars and it's for creating constants!

https://github.com/Mikolaj/horde-ad/blob/95fba8f42f5063d9599606a3308e7c5275d5e81c/src/HordeAd/Core/DualNumber.hs#L128-L129

Mikolaj:
I agree the ambiguity is awful and we avoid the clash with constant from

https://hackage.haskell.org/package/orthotope-0.1.2.0/docs/Data-Array-ShapedS.html#v:constant

by calling our constant dual number functions konst*, after package hmatrix, so your proposal makes sense. Another option is `fromPrimal`.

PR #30, opened 2022-04-28
Rename scalar to konst

tomjaguarpaw:
Closes https://github.com/Mikolaj/horde-ad/issues/29#issuecomment-1112388139

If merging, please delete the source branch.


Mikolaj:
Oh dear, I probably totally miscommunicated? I meant "constant" does not lead to confusion, because instead of "constant" we use "konst*" everywhere. So, I thought your proposal to use "constant" was spot on. :)

tomjaguarpaw:
OK, no problem. How about this?

Mikolaj (approved):
Perfect (if CI agrees).

Mikolaj:
Merging. Thank you!

PR #33, opened 2022-05-08
Reduce type params

tomjaguarpaw:
What do you think of this as a first attempt of reducing type parameters? Please merge if you like it.


Mikolaj (approved):
This looks great. Merging.

PR #35, opened 2022-05-09
Argument record example

tomjaguarpaw:
Here's an example of using more general argument structures, i.e. not simply one vector of each rank. What do you think?

### DIFF
--- a/src/HordeAd/Core/Again.hs
+++ b/src/HordeAd/Core/Again.hs
@@ -24,6 +24,7 @@ import Control.Monad.Trans.State
     runState,
   )
 import qualified Data.Array.ShapedS as OS
+import qualified Data.Biapplicative as B
 import Data.Functor.Identity (Identity (Identity), runIdentity)
 import Data.Kind (Type)
 import Data.List (foldl')
@@ -515,9 +516,14 @@ dLet ::
   m (Dual dual t)
 dLet = dLetS known
 
--- This is a Biapplicative, but we're not using that type class yet.
--- Perhaps we should.
 newtype ArgAdaptor s t pd = ArgAdaptor (State Int (DeltaMap s -> t, pd))
+
+instance B.Bifunctor (ArgAdaptor s) where
+  bimap = bimapArgAdaptor
+
+instance B.Biapplicative (ArgAdaptor s) where
+  bipure = pureArgAdaptor
+  f <<*>> x = B.bimap (uncurry ($)) (uncurry ($)) (liftB2 f x)
 
 runArgAdaptor ::
   ArgAdaptor s t pd ->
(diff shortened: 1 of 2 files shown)


src/HordeAd/Core/Again.hs
```diff
@@ -830,6 +856,28 @@ example2 =
...
+    (OS.Array '[p, q] s) ->
+  OS.Array '[l, q] s ->
+  ( OS.Array '[l, q] s,
+    ARecord
+      (OS.Array '[m, n] s)
+      (OS.Array '[n, p] s)
+      (OS.Array '[p, q] s)
+  )
+example2Record (ARecord x y z) t =
+  runDualMonadAdapt (B.bipure ARecord ARecord <<*>> adaptArg x <<*>> adaptArg y <<*>> adaptArg z) t bazRecord
```
tomjaguarpaw, line 879: Then you can get the gradient using this `Biapplicative` structure.  Further simplifications are possible but I wanted to know what you thought of this first.

Mikolaj (approved):
Definitely worth merging. Please do.

src/HordeAd/Core/Again.hs
tomjaguarpaw: I guess I don't really understand because even if the original argument is an unboxed storable vector of scalars you will have to break those scalars out into (unboxed, unstorable) `D`s for the entirety of the differentiated program. So, I'm can't grasp why this is a useful thing to do.

Anyway, it doesn't really matter if I can't grasp it, because if your version can do it then I'm sure my version can be persuaded to do it too, and we can investigate that when appropriate.

tomjaguarpaw: > That's why I don't store the collection of the initial dual numbers coming from parameters (the pairs of original value and the corresponding delta variable) in Ds, but in a pair of vectors

Hmm, isn't that what rank 1 is for?

Mikolaj: That's a good question. The only example for rank 0 I've seen is `prod` and it can be implemented on rank 1 as well. The `prod` benchmark could easily be extended with the rank 1 version but our benchmarks are broken anyway. I expect the rank 1 version to be much slower than rank 0 (with in-place updates), because vector indexing is slow (in-place updates are hard to implement there). However, perhaps we should just fix vector indexing, because it's so pervasive. There is a ticket about that, but it's outdated, because I can't get reliable benchmark data for it.

However, "pair of vectors" is beneficial even for rank 1 when you have many vectors and then "isn't that what rank 2 is for" may not apply when the vectors have different sizes. I don't remember how large the performance difference is in such cases (the vectors are both boxed, so it must be much less than for rank 0).

tomjaguarpaw: I see. There is a lot of complexity here that I am not familiar with.

PR #37, opened 2022-05-12
Weaken IsDualWithScalar for returnLet

tomjaguarpaw:
This avoids asking more than necessary of the callers of `returnLet`.


Mikolaj (approved):
This definitely makes sense even though, as explained in github discussion, the main restriction is elsewhere. But perhaps something can be done with the `Dual` restriction. However I'd need any tests first, or a concrete plan for a test, so that I don't generalize without a measure of success and a guard against future breakage.

Mikolaj:
Let me rebase this to `master` so that I can use this for `master` work, too.

Mikolaj:
Fixed and merged. Thank you!

PR #38, opened 2022-05-13
Delayed-outlined primitive numeric functions

### LINKED ISSUE #36: Try traces that contain unrolled syntax trees with many primitive functions instead of delta expressions
This is described in the google doc. That would amount to defining the big syntax tree datatype, copy-pasting DualNumber.hs and changing each (simple enough) function to the new scheme and then writing a short easy translation to the old machinery so that it can compute gradients or derivatives as before.

Mikolaj:
Closes #36.


tomjaguarpaw:
Interesting. I wasn't expecting it to be done like this. I was expecting "outlining" to be a new mode, alongside forward and reverse.

Mikolaj:
> Interesting. I wasn't expecting it to be done like this. I was expecting "outlining" to be a new mode, alongside forward and reverse.

Hah, that's precisely how I did it for the first half an hour. But I noticed I have to duplicate the delta expression grammars with small changes, replicate variable binding, state, etc., and I thought to myself, what if I just add to what's there instead of copying. The benefit is that delaying can now be done selectively.  The drawback is that the old system is slightly polluted with the new mechanism (the extra delta expression constructor for each rank) and that to delay every function that belongs to numeric classes, every non-numeric function application needs to be wrapped in `Out`, which is verbose for non-numeric portions of the code.

tomjaguarpaw:
Ah yes, you would have to add a whole new "parallel" grammar of functions.

Mikolaj:
Thank you for your comments. I've added yet another method of outlining, using laziness. A few other variants would make sense, but this should be enough for experimentation and users can then decide how to develop this further. Merging.

Mikolaj:
Yes, it's fully merged, so no point keeping it around. Let me give you admin rights so that you can tidy up as you see fit.

Mikolaj:
Doh, now I remember I tried and I can't, because we don't have an org, but it's my personal repo. Let me know if I'm missing a way to do that or if you can think of a good org to join/create.

PR #49, opened 2022-05-20
test suite parallelization

### LINKED ISSUE #40: cabal test takes ages to complete and uses only 2 cores
When running `cabal test` on my `8x Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz` with `32 GB RAM` I had to cancel it after some 18 minutes:

```
$ cabal test 
Build profile: -w ghc-9.2.2 -O1
In order, the following will be built (use -v for more details):
 - horde-ad-0.1.0.0 (lib) (configuration changed)
 - horde-ad-0.1.0.0 (lib:testLibrary) (configuration changed)
 - horde-ad-0.1.0.0 (test:shortTestForCI) (configuration changed)
 - horde-ad-0.1.0.0 (test:extremelyLongTest) (dependency rebuilt)
Configuring library for horde-ad-0.1.0.0..
Preprocessing library for horde-ad-0.1.0.0..
Building library for horde-ad-0.1.0.0..
Configuring library 'testLibrary' for horde-ad-0.1.0.0..
Preprocessing library 'testLibrary' for horde-ad-0.1.0.0..
Building library 'testLibrary' for horde-ad-0.1.0.0..
Configuring test suite 'shortTestForCI' for horde-ad-0.1.0.0..
Preprocessing test suite 'extremelyLongTest' for horde-ad-0.1.0.0..
Building test suite 'extremelyLongTest' for horde-ad-0.1.0.0..
[...]
Preprocessing test suite 'shortTestForCI' for horde-ad-0.1.0.0..
Building test suite 'shortTestForCI' for horde-ad-0.1.0.0..
Running 1 test suites...
Running 1 test suites...
Test suite extremelyLongTest: RUNNING...
^C
```

More importantly, this test suite appears to utilize just 2 CPU cores. It turns my laptop into a small cooker, though: `+96 °C` aka `+204.8 °F` (CPU temp.).

sfindeisen:
Just a couple of flags + README.

Fixes #40

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,8 @@ Running tests in parallel
 -------------------------
 
 ```
-cabal run shortTestForCI --enable-optimization -- +RTS -N -RTS
+$ cabal run shortTestForCI    --enable-optimization -- +RTS -N -RTS
+$ cabal run extremelyLongTest --enable-optimization -- +RTS -N -RTS
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -39,8 +39,7 @@ Running tests in parallel
 -------------------------
 
 ```
-$ cabal run shortTestForCI    --enable-optimization -- +RTS -N -RTS
-$ cabal run extremelyLongTest --enable-optimization -- +RTS -N -RTS
+cabal run shortTestForCI --enable-optimization -- +RTS -N -RTS
 ```
 
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,31 @@ ensures that the code is compiled with opimization and so executes the rather
 computation-intensive testsuite in reasonable time.
 
 
+Running tests
+-------------
+
+The test suite is run in parallel mode by default:
+
+```
+cabal test shortTestForCI    --enable-optimization
+cabal test extremelyLongTest --enable-optimization
+```
+
+However, this may cause extra printf messages from within the tests to be out of order. To keep your screen tidy, simply redirect `stderr`, e.g.: `2>/dev/null`:
+
+```
+cabal test shortTestForCI    --enable-optimization 2>/dev/null
+cabal test extremelyLongTest --enable-optimization 2>/dev/null
+```
+
+You can also run the test suite sequentially:
+
+```
+cabal test shortTestForCI    --enable-optimization -f test_seq
+cabal test extremelyLongTest --enable-optimization -f test_seq
+```
+
+
 Coding style
 ------------
 
--- a/README.md
+++ b/README.md
@@ -35,14 +35,6 @@ ensures that the code is compiled with opimization and so executes the rather
 computation-intensive testsuite in reasonable time.
 
 
-Running tests in parallel
--------------------------
-
-```
-cabal run shortTestForCI --enable-optimization -- +RTS -N -RTS
-```
-
-
 Coding style
 ------------
 
--- a/README.md
+++ b/README.md
@@ -35,6 +35,14 @@ ensures that the code is compiled with opimization and so executes the rather
 computation-intensive testsuite in reasonable time.
 
 
+Running tests in parallel
+-------------------------
+
+```
+cabal run shortTestForCI --enable-optimization -- +RTS -N -RTS
+```
+
+
 Coding style
 ------------
 
--- a/horde-ad.cabal
+++ b/horde-ad.cabal
@@ -99,7 +99,7 @@ common options
     cpp-options:      -DEXPOSE_INTERNAL
 
 common exe-options
-  ghc-options:        -rtsopts
+  ghc-options:        -threaded -rtsopts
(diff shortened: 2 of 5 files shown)


sfindeisen:
Fixes #40

Mikolaj:
Thank you. This works.

However, outputs of different nn training tests are mixed randomly. Perhaps we need a cabal flag that forbids outputting extra messages in tests? That flag would be used precisely for quick parallel runs. I even think the flag should by on by default and if somebody wants to see the details of a test, the flag can be turned off on commandline (would be great to describe this in README). However, if the flag is on by default, perhaps let's make parallelism of tests the default as well? I suppose `ghc-options: "-with-rtsopts=-N"` might do the trick.

I think benchmarks should not be run in parallel and neither a single benchmark should be allowed to use many cores, so let's add `-threaded` and `-N` only to tests.

Mikolaj:
I took the liberty of converting to draft (top left, just below reviewers), as I've learned how to do that yesterday, so I still remember. :)

Mikolaj:
Yes, I had in mind adding a cabal flag that is toggled on commandline. The drawback is that it doesn't work if somebody runs the test executable directly, without cabal. But there is no real reason to do that, I think.

But I like the stderr idea. Could you describe how to use it in README? And we'd need to factor out all our prints in test code and change switch to them to use stderr.

OTOH, doesn't tasty print *anything* to stderr, e.g., messages from failed tests? Don't messages about uncaught exceptions go to stderr by default? So the user would need to redirect that to /tmp/log, not /dev/null. I hope tasty at least prints "FAIL" on stdout and then the details can be looked up in /tmp/log.

sfindeisen:
I've just run `cabal run shortTestForCI --enable-optimization` and stderr is empty.

Mikolaj:
> I've just run `cabal run shortTestForCI --enable-optimization` and stderr is empty.

Despite the test errors you are getting? That's great news. In that case, only Haskell exceptions remain, which I checked are thrown to stderr.

sfindeisen:
Yes, I can see both test failures and custom printfs from within the code on stdout, like here:

```
Epochs to run/max batches per epoch: 1/1
[Epoch 1]
(Batch 1 with 5000 points)
Validation error: 88.65%
(Batch 3 with 5000 points)
Training error:   89.74%
Validation error: 13.11%
Validation error: 89.91%
(Batch 4 with 5000 points)
OK (113.37s)
    VV artificial 1 2 3 4 5 1 2 3 4 0 20 2421 5.0:             OK (9.93s)
    VV artificial 5 4 3 2 1 5 4 3 2 0 18 2393 1.0:             FAIL (45.60s)
      test/common/TestMnistFCNN.hs:198:
      expected: 0.7756000000000001
       but got: 0.7882
      Use -p '/VV artificial 5 4 3 2 1 5 4 3 2 0 18 2393 1.0/' to rerun this test only.
```

I will try to factor out all our prints in test code and switch them to use stderr.

sfindeisen:
> In that case, only Haskell exceptions remain, which I checked are thrown to stderr.

[Always](https://downloads.haskell.org/~ghc/latest/docs/html/users_guide/runtime_control.html#rts-flag--xc)?...

Mikolaj:
I assumed so, but only tried `error` in ghci...

Mikolaj (approved):
LGTM. Please feel free to either click "Merge pull request" to retain the meandering commit history of this PR or "Squash and Merge" to make the PR a single commit.

sfindeisen:
Thanks but it looks I do not have write access.

Mikolaj:
I've re-sent the invite.

Mikolaj:
My short test run finishes in 20min wall clock time in parallel and in 21min sequentially under moderate system load. This is probably due to me having only 4 cores and 8 threads, due to the tests and ML in general depending more on per-CPU (not per-core) cache and RAM speeds than CPU speed )and I don't have many CPU caches nor RAMs), and due to (I suspect) test groups that finished hogging a thread until the group that currently gets displayed finishes all tests (and we have many groups of differing running length). Let me try the long test.

Mikolaj:
And it may be due to the stop-the-world GC (this can be changed), not optimized GC settings and the fact that performance is currently broken (e.g., last time I measured, 3 times greater allocation than with older GHCs). So let's leave it as is and once performance is improved, parallelization should be a benefit.

sfindeisen:
On my machine (8 cores) under low system load:

```
$ cabal test shortTestForCI    --enable-optimization
...
13 out of 124 tests failed (1166.66s)
```

```
$ cabal test shortTestForCI    --enable-optimization -f test_seq
...
13 out of 124 tests failed (1368.99s)
```

which is an improvement of `15%`.

Mikolaj:
That's good enough and I improved the printouts so that they don't mix up so badly in parallel now.

PR #51, opened 2022-05-25
improve QuickCheck tests

### LINKED ISSUE #27: Improve QuickCheck tests
It would be great to factor out the testing machinery that checks consistency of all the kinds of derivatives and gradients and all kinds of computing the value of the objective function.

Ideally, we'd also randomly generate the object function and run the machinery on that. No idea if the consistency conditions hold for non-continuous functions, but perhaps they do for non-differentiable ones.

Some part of this ticket are relatively easy or at least don't require an intimation knowledge of the library, so I'm marking it as a good first issue. Even partial solutions would be very helpful.

Edit: Initially I thought shrinking needs adjustment as well, but it seems it's correct as of now. If we randomly generate functions and parameters that match their domains, it would be important that the functions and the parameters are shrunk in tandem to maintain the match.

sfindeisen:
Fixes #27

### DIFF
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -1,0 +1,369 @@
+{-# LANGUAGE DataKinds, RankNTypes, TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+module TestCommon (dReverse0, dReverse1, fX, fX1Y, fXXY, fXYplusZ, fXtoY, freluX,
+                   vec_omit_scalarSum_aux, sumElementsV, altSumElementsV,
+                   sinKonst, sinKonstOut, sinKonstDelay, sinKonstS, sinKonstOutS, sinKonstDelayS,
+                   powKonst, powKonstOut, powKonstDelay,
+                   listsToParameters,
+                   qcTest, fquad, quad,
+                   atanReadmeM, atanReadmeDReverse,
+                   vatanReadmeM, vatanReadmeDReverse,
+                  ) where
+
+import Prelude
+
+import qualified Data.Array.DynamicS as OT
+import qualified Data.Array.ShapedS as OS
+import qualified Data.Strict.Vector as Data.Vector
+import qualified Data.Vector.Generic as V
+import           Numeric.LinearAlgebra (Vector)
+import qualified Numeric.LinearAlgebra as HM
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+import HordeAd hiding (sumElementsVectorOfDual)
+import HordeAd.Core.DualClass (IsPrimal, dAdd, dScale)
+
+-- Unfortunately, monadic versions of the operations below are not
+-- polymorphic over whether they operate on scalars, vectors or other types,
+-- so we should probably abandon them.
+
+(+\) :: DualMonad d r m
+     => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(+\) u v = returnLet $ u + v
+
+(*\) :: DualMonad d r m
+     => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(*\) u v = returnLet $ u * v
+
+(**\) :: DualMonad d r m
+      => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(**\) u v = returnLet $ u ** v
+
+dReverse0
+  :: HasDelta r
+  => (DualNumberVariables 'DModeGradient r
+      -> DualMonadGradient r (DualNumber 'DModeGradient r))
+  -> [r]
+  -> ([r], r)
+dReverse0 f deltaInput =
+  let ((results, _, _, _), value) =
+        dReverse 1 f (V.fromList deltaInput, V.empty, V.empty, V.empty)
+  in (V.toList results, value)
+
+fX :: DualMonad 'DModeGradient Float m
+   => DualNumberVariables 'DModeGradient Float
+   -> m (DualNumber 'DModeGradient Float)
+fX variables = do
+  let x = var0 variables 0
+  return x
+
+fX1Y :: DualMonad 'DModeGradient Float m
+     => DualNumberVariables 'DModeGradient Float
+     -> m (DualNumber 'DModeGradient Float)
+fX1Y variables = do
+  let x = var0 variables 0
+      y = var0 variables 1
+  x1 <- x +\ 1
+  x1 *\ y
+
+fXXY :: DualMonad 'DModeGradient Float m
+     => DualNumberVariables 'DModeGradient Float
+     -> m (DualNumber 'DModeGradient Float)
+fXXY variables = do
+  let x = var0 variables 0
+      y = var0 variables 1
+  xy <- x *\ y
+  x *\ xy
+
+fXYplusZ :: DualMonad 'DModeGradient Float m
+         => DualNumberVariables 'DModeGradient Float
+         -> m (DualNumber 'DModeGradient Float)
+fXYplusZ variables = do
+  let x = var0 variables 0
+      y = var0 variables 1
+      z = var0 variables 2
+  xy <- x *\ y
+  xy +\ z
+
+fXtoY :: DualMonad 'DModeGradient Float m
+      => DualNumberVariables 'DModeGradient Float
+      -> m (DualNumber 'DModeGradient Float)
+fXtoY variables = do
+  let x = var0 variables 0
+      y = var0 variables 1
+  x **\ y
+
+freluX :: DualMonad 'DModeGradient Float m
+       => DualNumberVariables 'DModeGradient Float
+       -> m (DualNumber 'DModeGradient Float)
+freluX variables = do
+  let x = var0 variables 0
+  reluAct x
+
+quad :: DualMonad d r m
+     => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+quad x y = do
+  x2 <- returnLet $ square x
+  y2 <- y *\ y
+  tmp <- x2 +\ y2
+  tmp +\ 5
+
+fquad :: forall r d m. DualMonad d r m
+      => DualNumberVariables d r -> m (DualNumber d r)
+fquad variables = do
+  let x = var0 variables 0
+      y = var0 variables 1
+  quad x y
+
+vec_omit_scalarSum_aux
+  :: DualMonad d r m
+  => DualNumberVariables d r -> m (DualNumber d r)
+vec_omit_scalarSum_aux vec = returnLet $ foldlDual' (+) 0 vec
+
+sumElementsV
+  :: DualMonad d r m
+  => DualNumberVariables d r -> m (DualNumber d r)
+sumElementsV variables = do
+  let x = var1 variables 0
+  returnLet $ sumElements0 x
+
+altSumElementsV
+  :: DualMonad d r m
+  => DualNumberVariables d r -> m (DualNumber d r)
+altSumElementsV variables = do
+  let x = var1 variables 0
+  returnLet $ altSumElements0 x
+
+-- hlint would complain about spurious @id@, so we need to define our own.
+id2 :: a -> a
+id2 x = x
+
+sinKonst
+  :: DualMonad d r m
+  => DualNumberVariables d r -> m (DualNumber d r)
+sinKonst variables = do
+  let x = var1 variables 0
+  return $ sumElements0 $
+    sin x + (id2 $ id2 $ id2 $ konst1 1 2)
+
+sinKonstOut
+  :: ( DualMonad d r m
+     , Floating (Out (DualNumber d (Vector r))) )
+  => DualNumberVariables d r -> m (DualNumber d r)
+sinKonstOut variables = do
+  let x = var1 variables 0
+  return $ sumElements0 $
+    unOut $ sin (Out x) + Out (id2 $ id2 $ id2 $ konst1 1 2)
+
+sinDelayed :: (Floating a, IsPrimal d a) => DualNumber d a -> DualNumber d a
+sinDelayed (D u u') = delayD (sin u) (dScale (cos u) u')
+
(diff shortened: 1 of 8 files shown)


sfindeisen:
Work in progress. I implemented `TestCommon.hs` with `qcTest` and other common functions.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,369 @@
+{-# LANGUAGE DataKinds, RankNTypes, TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
+module TestCommon (dReverse0, dReverse1, fX, fX1Y, fXXY, fXYplusZ, fXtoY, freluX,
```
Mikolaj, line 3: fX, fX1Y, fXXY, fXYplusZ, fXtoY, freluX and probably a lot of others are not really auxiliary, but one-time silly tests. However, no harm factoring them out, using some of them in more than one test (but, really, most of them are silly and trivial), making sure nothing is repeated, inlining again everything that is only used in one place.

sfindeisen: I was expecting that.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,413 @@
...
+import qualified Numeric.LinearAlgebra as HM
+import           Test.Tasty
+import           Test.Tasty.QuickCheck
+
+import HordeAd hiding (sumElementsVectorOfDual)
+import HordeAd.Core.DualClass (IsPrimal, dAdd, dScale)
+
+-- Unfortunately, monadic versions of the operations below are not
+-- polymorphic over whether they operate on scalars, vectors or other types,
+-- so we should probably abandon them.
```
Mikolaj, line 30: Hah, this whole comment is, fortunately, no longer true and can be removed (and in another file, unless it's been factored out).

Mikolaj:
Splendid. Will review in a couple of days.

And is the whole idea of the test mathematically clear: derivative (calculated in two ways) vs gradient vs perturbation? Is it documented well enough in the test and in the functions

https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Core/Engine.hs#L144

https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Core/Engine.hs#L180

https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Core/Engine.hs#L217

that are the main actors of the test (their haddocks could refer to the test as both their specification and test). I can only find an equation from a different context and so less clear

https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Internal/Delta.hs#L331

but even this haddock could refer to the test, I think.

sfindeisen:
I suggest we make documenting those 3 functions (`dReverse`, `dForward`, `dFastForward`) a separate ticket. I will be happy to work on that too (at some point). OK to create?

sfindeisen:
https://github.com/Mikolaj/horde-ad/issues/52

Mikolaj:
> I suggest we make documenting those 3 functions (`dReverse`, `dForward`, `dFastForward`) into a separate ticket. I will be happy to work on that too (at some point). OK to create?

Perfect. Thank you.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,179 @@
...
+(*\) :: DualMonad d r m
+     => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(*\) u v = returnLet $ u * v
+
+(**\) :: DualMonad d r m
+      => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(**\) u v = returnLet $ u ** v
+
+-- Checks if 2 numbers are close enough.
+close1 :: Double -> Double -> Bool
```
Mikolaj, line 34: Since we are generalizing, could we generalize this to arbitrary `Num r`? And similarly everywhere else in this file. We need a concrete type as soon as we actually test, but in test tools, restriction to Double only adds noise.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,179 @@
...
+                         , Floating (Out (DualNumber d (Vector r)))
+                         , Floating (Out (DualNumber d (OS.Array '[2] r))) )
+           => DualNumberVariables d r -> m (DualNumber d r))
+       -> ((Double, Double, Double) -> ([Double], [Double], [Double], [Double]))
+       -> TestTree
+quickCheckTest0 txt f fArg =
+  qcTestRanges txt f (listsToParameters4 . fArg) ((-2, -2, -2), (2, 2, 2)) ((-1e-7, -1e-7, -1e-7), (1e-7, 1e-7, 1e-7)) (-10, 10)
+
+-- A quick check to compare the derivatives and values of 2 given functions.
+cmpTwo :: (forall d r m. ( DualMonad d r m
```
Mikolaj, line 78: If we generalize from `Double` to `Num r` or `Floating r`, we can probably merge the two foralls and move them to the start of this type signature.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,179 @@
...
+       -> Domains Double
+       -> Domains Double
+       -> Property
+cmpTwoSimple f1 f2 parameters ds =
+  cmpTwo f1 f2 parameters parameters ds ds
+
+-- A quick consistency check of all the kinds of derivatives and gradients
+-- and all kinds of computing the value of the objective function.
+qcPropDom :: (forall d r m. ( DualMonad d r m
+                         , r ~ Double
```
Mikolaj, line 116: broken formatting again?

test/common/TestCommon.hs
```diff
@@ -0,0 +1,179 @@
...
+      => DualNumber d r -> DualNumber d r -> m (DualNumber d r)
+(**\) u v = returnLet $ u ** v
+
+-- Checks if 2 numbers are close enough.
+close1 :: Double -> Double -> Bool
+close1 a b = abs (a - b) <= 1e-4
+
+-- Checks if 2 number pairs are close enough.
+close2 :: (Double, Double) -> (Double, Double) -> Property
+close2 (a1, b1) (a2, b2) = (close1 a1 a2) .&&. (close1 b1 b2)
```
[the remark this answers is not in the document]
sfindeisen: Hey, why so much parenfobia here?... Those parens are in love. Seriously, can we agree that some parens can actually improve code readability for a human, even if they are spurious? Or is this a fascist project with `hlint` being the rule?

In any case, those parens must go in pairs, like `(` and `)`. The `)` and `(` variant, or triples, is not yet supported by the (parenfobic) GHC. With the help of our generous sponsors we are currently coordinating a spontaneous worldwide protest to cancel that.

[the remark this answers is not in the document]
Mikolaj: I'm not qualified to judge, but I suspect the two mechanisms have benefits of similarly modest size: the first being parens that make grouping more explicit and so easier to read and the second being a uniform rule for where parens are written (current rule: as few as possible) that help the brain guess correct parsing from deterministic parens.

Given that we don't have an alternative uniform rule for parens, we have to choose between the two benefits. Given that, additionally, writing least parens produces less hlint noise and it seems to be more common in the community, I'd opt for this course of action. What do you think?

test/common/TestCommon.hs
sfindeisen: Throughout the codebase I have seen different ways of formatting this kind of stuff, for example [here](https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Core/Engine.hs#L246) or [here](https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/test/common/TestSimpleDescent.hs#L64) or [here](https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/test/common/TestSimpleDescent.hs#L185). What is the right way? It is enough if you point me to the right example.

test/common/TestCommon.hs
```diff
@@ -0,0 +1,179 @@
+{-# LANGUAGE DataKinds, RankNTypes, TypeFamilies #-}
+{-# OPTIONS_GHC -Wno-incomplete-uni-patterns #-}
```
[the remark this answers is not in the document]
sfindeisen: Fixed

test/common/TestCommon.hs
sfindeisen: A simple fix like that doesn't seem to work, due to existing `r ~ Double` or similar bindings, e.g. here: https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/test/common/TestMnistCNN.hs#L623 .

test/common/TestCommon.hs
Mikolaj: There's lots of broken formatting in this codebase from blind query-replace. I'm sorry about that. The rule is that `=>` and `->` align with `::`, unless they are inside parens, in which case they start in the column one higher than where the starting paren is. So all the three examplex are fine. The first breaks line after a long function name and starts the new line with `::`, the second doesn't break line after the forall phrase, the third is the "inside parens" case. Another rule is that if a line starts with a comma, it aligns with the paren above and is followed by a space, just as the paren.

test/common/TestCommon.hs
Mikolaj: You are right, this is not trivial and out of scope of this PR. I had to do the following:

```hs
import           HordeAd.Core.DualClass (Dual)

cmpTwo
  :: ( m ~ DualMonadForward r, d ~ 'DModeDerivative, Dual d r ~ r
     , DualMonad d r m )
  => (DualNumberVariables d r -> m (DualNumber d r))
  -> (DualNumberVariables d r -> m (DualNumber d r))
  -> Domains r
  -> Domains r
  -> Domains r
  -> Domains r
  -> Property

cmpTwoSimple
  :: ( m ~ DualMonadForward r, d ~ 'DModeDerivative, Dual d r ~ r
     , DualMonad d r m )
  => (DualNumberVariables d r -> m (DualNumber d r))
  -> (DualNumberVariables d r -> m (DualNumber d r))
  -> Domains r
  -> Domains r
  -> Property
```

test/common/TestCommon.hs
sfindeisen: Yet a separate case are perhaps short argument names, like [here](https://github.com/Mikolaj/horde-ad/blob/91dc3840ee27cc75ce1b7c5a1411600eb66bfe94/src/HordeAd/Core/Engine.hs#L301)? Did you think of enforcing those rules on the server side as part of CI/CD, or of implementing them as part of `haskell-mode`?

test/common/TestCommon.hs
sfindeisen: Gladly imported, but in the other cases this didn't work. Shall we have a ticket to similarly simplify some of those type signatures?

test/common/TestCommon.hs
Mikolaj: You are right, the rule should read "`=>` and `->` that start a line align...". Yes, I think the new haskell mode (not the decade old version I'm using) may be customizable enough to support such a rule. Or you can just hack some emacs-lisp. :) A really tempting alternative is ormolu or another auto-formatter. It prevents formatting getting broken due to `sed` query-replace. However, I have yet to see auto-formatting that doesn't make me cringe and also auto-formatters can't cope with 80-characters rule, in part because the rule has to admit exceptions, e.g., for very long strings (or you need to refactor code to split the strings; auto-formatter can't do that).

test/common/TestCommon.hs
```diff
@@ -0,0 +1,172 @@
...
+         .&&. ffValue == revValue
+         -- Gradients and derivatives agree.
+         .&&. close1 (dt * derivative)
+                     (dotParameters gradient ds)
+         -- Objective function value is unaffected by perturbation.
+         .&&. ffValue == valueAtPerturbation
+         -- Derivative approximates the perturbation of value.
+         .&&. close1 (primalValue
+                                  f (addParameters
+                                                   args perturbation))
```
Mikolaj, line 134: This is a completely borked indentation from the `sed` refactorings I was doing. Please don't bother fixing. I will do at some point.

test/common/TestMnistCNN.hs
```diff
@@ -585,24 +587,11 @@ mnistCNNTestsLong = testGroup "MNIST CNN long tests"
...
-           .&&. dForward f parameters ds === ff
-           .&&. closeEq (dfDot f parameters ds) ff
-           .&&. dForward fP parameters ds === ffP
-           .&&. closeEq (dfDot fP parameters ds) ffP
-           .&&. close (primalValue fP (addParameters parameters
-                                                     parametersPerturbation))
-                      (ffPValue + fst perturbedffP)
+        in
+            qcPropDom f  parameters ds parametersPerturbation 1 .&&.
+            qcPropDom fP parameters ds parametersPerturbation 1 .&&.
```
Mikolaj, line 592: A task for a separate PR (perhaps the one where we link the defining properties of gradients/derivatives to the Quickcheck properties in tests) is to also randomly generalte the `dt` values that arefixed at `1` here. They are fixed, because in an older version of the code, they were fixed at `1` somewhere deep in the internals of the library. But now they are exposed so we can vary them in tests.

Mikolaj:
Oh, I see you squash-merged. Sure, whatever suits.

sfindeisen:
Well, I thought this is preferred? Like in the last one? Or what is the rule?

Mikolaj:
All three methods are useful. The only rule is that in the "Create a merge commit" method, the branch should be rebased beforehand to keep the history not tangled (the other two methods effectively rebase automatically). You are right, you squashed in #49, while I merged in #39. BTW, `gitk` shows the colourful commit history quite well.

Issue #48, opened 2022-05-20
Rewrite one of shaped tests not to forget type-level shapes of parameters at each gradient descent update


Mikolaj:
Edit: #66 willl probably solve this ticket and the only thing left will be re-implementing all big examples in the new style.

Create a variant of one of the `*Shaped.hs` examples in https://github.com/Mikolaj/horde-ad/tree/master/src/HordeAd/Tool in the spirit of `fquadRecord` from https://github.com/Mikolaj/horde-ad/blob/master/test/common/TestSimpleDescent.hs, which was in turn modelled after @tomjaguarpaw 's https://github.com/Mikolaj/horde-ad/blob/58dd1a7a348412472a29d4144d51f76e2177f54e/src/HordeAd/Core/Again.hs#L878-L879

The problem we are trying to solve is that in the original `*Shaped.hs` examples the shaped tensors are stored in the tuple of vectors of parameters as untyped tensors, with all shape information wiped out. In particular, our typing does not ensure that each step of gradient descent does not change the shape of the tensor parameters. In other words, currently, if we changed the implementation in a way that perturbs the shapes while they are stored, our types would not catch this violation of the informal invariant.

The `fquadRecord` example from https://github.com/Mikolaj/horde-ad/blob/master/test/common/TestSimpleDescent.hs solves this problem, but on a small scale and with lots of boilerplate. The hope is, after implementing this in a bigger example, we'd see how to perform this stronger typing in general and we'd capture some of the boilerplate as reusable functions. A possible outcome is a realization that the current formalism is not adequate for capturing the boilerplate, unlike Tom's formalism, and so should try to converge towards Tom's. In particular, it may turn out that `Data.Biapplicative` adds value, though let's not assume this, but find out and demonstrate how the need arises and what the alternatives are.

Issue #42, opened 2022-05-17
Implement bachnorm layer and add it to some tests


Mikolaj:
https://en.wikipedia.org/wiki/Batch_normalization
https://penkovsky.com/neural-networks/day4/

I'm sure there are good blog posts about that. Looking this up in pytorch and/or tensorflow API may help, too.

Perhaps adding this to the existing tests with convolutional and recurrent neural networks would be a good idea. We could measure how the training improves. With batchnorm perhaps we could switch activation function to relu, in examples where it's not already relu, and not lose convergence. Probably normalizing data from -1 to 1 (zero mean, unit variance) would also be useful for relu, so a convenient tools for normalizing is a related task (probably batchnorm is just much more general, e.g., it needs to be differentiable).

Two other keywords that may be related or may be even the same thing as one of the above, so worth googling:

* regularization functions, to keep params not too small or big

* activity_regularizer: Regularizer function applied to the output of the layer (its "activation")

Mikolaj:
A TODO left over from an experimental implementation of the algorithms in https://arxiv.org/pdf/2202.08587.pdf mentions similar kewords. When this ticket is closed, let's create a new one that mentions the parts of the TODO not completed and clarifies the terms used.

> Consider weight decay instead of normalization/regularization/gradient clipping. Verify if it's on by default in pytorch or tensorflow. When the forward gradient paper is published with source code, verify what they have used to avoid gradient explosion. If nothing, investigate why I'm quikly getting NaN in mnistTestCase2D 1024 1024 test and they don't.

Mikolaj:
Related transcript from datahaskell/lobby (https://matrix.to/#/!JSHTXUPHaIJhjEINXX:matrix.org?via=m.topoi.dev&via=gitter.im&via=matrix.org) follows the line:

---

Man of Letters

Hi! I'm implementing yet another automatic differentiation library with a bunch of co-conspirators and a major struggle in that is wrapping my head around some data science concepts, e.g., data normalization (https://github.com/Mikolaj/horde-ad/issues/42).

anton_5

Have you seen this paper https://arxiv.org/pdf/1502.03167.pdf ? It’s basically Algorithm 1 on page 3 and Algorithm 2 on page 4.
Man of Letters

Man of Letters

Oh, thanks, well explained, definitely easier to follow than source code of tensorflow or pytorch --- BTW, is that exactly what they are implementing as batch normalization? Things could have changed since 2015...

anton_5

[...]

PR #53, opened 2022-05-31
Update README.md

awf:
Some proposed changes to README.  Food for discussion, and partially an aspirational direction of travel for ergonomics.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,18 +1,121 @@
-# horde-ad
-Higher Order Reverse Derivatives Efficiently - Automatic Differentiation library based on the paper "Provably correct, asymptotically efficient, higher-order reverse-mode automatic differentiation" by Faustyna Krawiec, Neel Krishnaswami, Simon Peyton Jones, Tom Ellis, Andrew Fitzgibbon and Richard Eisenberg. Mature Haskell libraries with similar capabilities, but varying efficiency, are https://hackage.haskell.org/package/ad and https://hackage.haskell.org/package/backprop. See also https://github.com/Mikolaj/horde-ad/blob/master/CREDITS.md.
+# HORDE-AD: Higher Order Reverse Derivatives Efficiently 
 
-This is an early prototype, both in terms of the engine performance and the API and toolbox for the library user. The user should be ready to add missing primitives, as well as obvious tools that should be predefined but aren't. One can already build with the library all basic neural network architectures, such as fully connected, recurrent, convolutional and residual networks. The library, due to its loose coupling of differentiation and data containers, can naturally handle exotic variants of such networks that wouldn't express well in a language of matrices or tensors. Applications outside machine learning are plausible, given that the notion of a neural network is not hardwired into the formalism, but ad hoc redefined from basic blocks of general automatic differentiation whenever it's needed, e.g., in the tests and benchmarks of the library.
+This is an Automatic Differentiation library based on the paper _"Provably correct, asymptotically efficient, higher-order reverse-mode automatic differentiation"_ by Krawiec, Krishnaswami, Peyton Jones, Ellis, Fitzgibbon, and Eisenberg.
 
-Here is an example of computing the gradient of a function that goes from `R^3` to `R^2`:
+This is an early prototype, both in terms of the engine performance and the API and toolbox for the library user. The user should be ready to add missing primitives, as well as obvious tools that should be predefined but aren't. One can already build with the library all basic neural network architectures, such as fully connected, recurrent, convolutional and residual networks. The library, due to its loose coupling of differentiation and data containers, can naturally handle exotic variants of such networks that wouldn't express well in a language of matrices or tensors. 
 
-https://github.com/Mikolaj/horde-ad/blob/40951dab17ae90c5ce528a59216a39a3727eccb5/test/common/TestSingleGradient.hs#L442-L516
+Applications outside machine learning are very much in scope, given that the notion of a neural network is not hardwired into the formalism, but ad hoc redefined from basic blocks of general automatic differentiation whenever it's needed.
 
-Elsewhere in the same file is a computation of the forward derivative of the function and a QuickCheck test relating it to the gradient. It uses the same definition of the objective function and the same glue code for grouping parameters, etc. The ratio of signal to noise (maths to glue code) is much higher in more complex functions, e.g., neural networks. Several sample neural networks for MNIST digit classification are gathered in
+Mature Haskell libraries with similar capabilities, but varying efficiency, are https://hackage.haskell.org/package/ad and https://hackage.haskell.org/package/backprop. See also https://github.com/Mikolaj/horde-ad/blob/master/CREDITS.md.  
+The benchmarks at TODO show that this library has performance highly competitive with (i.e. faster than) those and PyTorch on CPU.  
+It is hoped that the separation of AD logic from matrix and tensor manipulation (deferred to [hmatrix] and [orthotope]) will enable similar speedups on numerical accelerators.
+
+## Computing the derivative of a simple function
+
+Here is an example of computing the gradient of a function that goes from `R^3` to `R`:
+
+```Haskell
+-- A function that goes from `R^3` to `R`. 
+-- [Arguments are tupled, could just as easily be curried. -awf]
+foo :: RealFloat a => (a,a,a) -> a
+foo (x,y,z) = 
+  let w = x * sin y 
+  in atan2 z w + z * w  -- [note extra w, we *want* to require returnLet up front --awf] 
+```
+
+In order to compute the gradient, we transform the function in the classic manner for forward-mode AD, to take `DualNumber` variables, 
+and the `DualMonad` which implements the system in the paper
+```Haskell
+fooM :: DualMonad d r m 
+     => DualNumberVariables d r -> m (DualNumber d r) 
+fooM variables = do { 
+     let x : y : z : _ = vars variables 
+     w <- x * sin y 
+     returnLet $ (atan2 z w + z * w)  -- [note extra w, we *want* to require returnLet up front --awf] 
+   }
+```
+We hope to automate this transfomation soon, automatically generating `fooM` from `foo`, but for now it's a mechanical, if somewhat tedious, manual task.
+
+And then the gradient of `foo` is
+```Haskell
+grad_foo :: HasDelta r => Domain0 r -> (Domain0 r, r) 
+grad_foo ds = 
+   let ((result, _, _, _), value) = 
+         dReverse 1 fooM (ds, V.empty, V.empty, V.empty) 
+   in (result, value) 
+```
+
+As can be verified by computing the gradient at `(1.1,2.2,3.3)`:
+```Haskell
+>>> grad_foo (mkDual [1.1, 2.2, 3.3])
+[7.75, 1.23, 5.52]
+```
+
+## Computing Jacobians.
+Now let's consider a function from 'R^n -> R^m'.  We don't want the gradient, but instead the Jacobian.
+```Haskell
+-- A function that goes from `R^3` to `R^2`. 
+foo :: RealFloat a => (a,a,a) -> (a,a)
+foo (x,y,z) = 
+  let w = x * sin y 
+  in (atan2 z w, z * w)
+```
+TODO: show how the 2x3 Jacobian emerges from here
+
+# Forall shapes and sizes
+
+An additional feature of this library is a type system for tensor shape arithmetic.  Look at this beauty.
+```Haskell
+convMnistTwoS x                                 -- input images, shape (batch_size, c_in, h, w)
+              ker1 bias1                        -- layer1 kernel, shape (c_out, c_in, kh+1, kw+1); and bias, shape (c_out)
+              ker2 bias2                        -- layer2 kernel, shape (c_out, c_out, kh+1, kw+1); and bias, shape (c_out)
+              weigthsDense                      -- dense weights shape (n_hidden, c_out, ((h+kh)/2 + kh)/2, ((w+kw)/2 + kw)/2) 
+              biasesDense                       -- dense biases shape (n_hidden)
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,19 +1,122 @@
...
+The benchmarks at TODO show that this library has performance highly competitive with (i.e. faster than) those and PyTorch on CPU.  
+It is hoped that the separation of AD logic from matrix and tensor manipulation (deferred to [hmatrix] and [orthotope]) will enable similar speedups on numerical accelerators.
+
+## Computing the derivative of a simple function
+
+Here is an example of computing the gradient of a function that goes from `R^3` to `R`:
+
+```Haskell
+-- A function that goes from `R^3` to `R`. 
+-- [Arguments are tupled, could just as easily be curried. -awf]
```
awf, line 19: These [-awf] comments are just for the pull request, so I didn't forget to make them.  They would be removed before merging.

Mikolaj:
Thank you so much. Please take your time.

README.md
```diff
@@ -1,19 +1,122 @@
...
+In order to compute the gradient, we transform the function in the classic manner for forward-mode AD, to take `DualNumber` variables, 
+and the `DualMonad` which implements the system in the paper
+```Haskell
+fooM :: DualMonad d r m 
+     => DualNumberVariables d r -> m (DualNumber d r) 
+fooM variables = do { 
+     let x : y : z : _ = vars variables 
+     w <- x * sin y 
+     returnLet $ (atan2 z w + z * w)  -- [note extra w, we *want* to require returnLet up front --awf] 
+   }
```
awf, line 35: Probably incorrect on every line - I am guessing, but I am fairly confident this should contain a `do`, a `<-` and a `returnLet`
I assume I should have written `sinM` and `atan2M` or similar?

README.md
```diff
@@ -1,19 +1,122 @@
...
+```
+We hope to automate this transfomation soon, automatically generating `fooM` from `foo`, but for now it's a mechanical, if somewhat tedious, manual task.
+
+And then the gradient of `foo` is
+```Haskell
+grad_foo :: HasDelta r => Domain0 r -> (Domain0 r, r) 
+grad_foo ds = 
+   let ((result, _, _, _), value) = 
+         dReverse 1 fooM (ds, V.empty, V.empty, V.empty) 
+   in (result, value)
```
awf, line 45: Similarly - whatever is correct here.

README.md
```diff
@@ -1,19 +1,122 @@
...
+grad_foo ds = 
+   let ((result, _, _, _), value) = 
+         dReverse 1 fooM (ds, V.empty, V.empty, V.empty) 
+   in (result, value) 
+```
+
+As can be verified by computing the gradient at `(1.1,2.2,3.3)`:
+```Haskell
+>>> grad_foo (mkDual [1.1, 2.2, 3.3])
+[7.75, 1.23, 5.52]
```
awf, line 51: This should remain a 2-liner, with as little line noise as possible. 
E.g. `grad_foo (DualVector.fromList [1.1 :: Float, 2.2, 3.3]`)

README.md
```diff
@@ -1,19 +1,122 @@
...
-
-Here is an example of computing the gradient of a function that goes from `R^3` to `R^2`:
-
-https://github.com/Mikolaj/horde-ad/blob/40951dab17ae90c5ce528a59216a39a3727eccb5/test/common/TestSingleGradient.hs#L442-L516
-
-Elsewhere in the same file is a computation of the forward derivative of the function and a QuickCheck test relating it to the gradient. It uses the same definition of the objective function and the same glue code for grouping parameters, etc. The ratio of signal to noise (maths to glue code) is much higher in more complex functions, e.g., neural networks. Several sample neural networks for MNIST digit classification are gathered in
+# HORDE-AD: Higher Order Reverse Derivatives Efficiently 
+
+This is an Automatic Differentiation library based on the paper _"Provably correct, asymptotically efficient, higher-order reverse-mode automatic differentiation"_ by Krawiec, Krishnaswami, Peyton Jones, Ellis, Fitzgibbon, and Eisenberg.
```
awf, line 3: I think we should add something like:

> Primarily implemented by @Mikolaj directly from the paper, and then subsequently in collaboration with the authors

Mikolaj: Or we can skip myself (I'm credited in CREDITS.md). Regardless, mentioning the paper authors would be great, as it immediately elevates the library from the category of fan fiction to the category of a canonical piece.

BTW, it doesn't really matter, but it was implemented directly from Simon's talk (with snippets of Haskell code) and only a few details Simon omitted were later taken from the paper (which explains some of the cases where terminology not always follows the paper). However, by transitivity, it was from the paper.

README.md
```diff
@@ -1,19 +1,122 @@
...
+```Haskell
+convMnistTwoS
+  :: forall kh kw num_hidden c_out h w c_in batch_size d r m.
+     ( KnownNat kh, KnownNat kw, KnownNat num_hidden, KnownNat c_out, KnownNat h, KnownNat w, KnownNat batch_size
+     , c_in ~ 1 -- greyscale
+     , 1 <= kh  -- kh greater than 1
+     , 1 <= kw  -- kw greater than 1
+     , DualMonad d r m )
+  => OS.Array '[batch_size, c_in, h, w] r
+  -> DualNumber d (OS.Array '[ c_out, c_in , kh + 1, kw + 1 ] r)
```
Mikolaj, line 95: To fit such things into a single line, perhaps the `DualNumber` type constructor could be renamed to `D` (the same as the value constructor of that type). That's what @tomjaguarpaw did in his variant of the implementation. `DualNumber` is less cryptic at the first glance, but more noisy in the long run, as we can see here.

README.md
```diff
@@ -1,19 +1,122 @@
...
+  -> DualNumber d (OS.Array '[c_out] r)
+  -> DualNumber d (OS.Array '[ c_out, c_out, kh + 1, kw + 1 ] r)
+  -> DualNumber d (OS.Array '[c_out] r)
+  -> DualNumber d (OS.Array '[ num_hidden, c_out,
+                                 GHC.TypeLits.* ((h + kh) `Div` 2 + kh) `Div` 2
+                                 GHC.TypeLits.* ((w + kw) `Div` 2 + kw) `Div` 2 -- [buglet? kh->kw? -awf]
```
Mikolaj, line 101: Yes, this is a bug and it was not caught by GHC due to the not implemented `transpose2` TODO above (now it's caught). This beautifully undermines the whole story of correctness of shapes guaranteed by dependent types.

The flop was possible, because Haskell is not a dependently typed language, as Agda or Idris, but a wide-spectrum language, where you can mix dependent and normal typing. Unfortunately, it's not easy to detect if such mixing is taking place or not. The new `transpose2S` function is fully dependently typed, while the `from2S` used before is, in fact, a coercion of matrices to rank 2 tensors, so its very purpose is to override dependent typing.

In the code you cited, after the TODO is fixed, the only remaining lax function is `reshapeS` that, at most, could guarantee on the type level that the number of elements does not change after the tensor reshaping. However what the resulting ranks are is unknown and so the function is a coercion and can hide bugs. We could instead use function `flattenS1S` that would require that the resulting tensor is of rank 1 and so no flexibility would be left in the output of the function. That's probably worth doing just for the sake of this example. However, at some point we need to decide how many differentiable tensor manipulation operations we want to maintain and if it's not better, e.g.,  to constrain the use of `reshapeS` with `:: DualNumber d (OS.Array '[... * ... * ...])` instead of adding `flattenS1S` that does the very same thing but in one word.

README.md
```diff
@@ -1,19 +1,122 @@
...
+Several sample neural networks for MNIST digit classification are gathered in
 
 https://github.com/Mikolaj/horde-ad/tree/master/src/HordeAd/Tool
 
 Among them are recurrent, convolutional and fully connected networks based on fully typed tensors (sizes of all dimensions are tracked in the types) as well as weakly typed fully connected networks built with, respectively, matrices, vectors and raw scalars (most flexible but slowest).
 
 
+# Derivative checking
+
+Elsewhere in the same file is a computation of the forward derivative of the function and a QuickCheck test relating it to the gradient. It uses the same definition of the objective function and the same glue code for grouping parameters, etc. The ratio of signal to noise (maths to glue code) is much higher in more complex functions, e.g., neural networks.
```
Mikolaj, line 118: This could be refined once we settle on the text above it, because by this point you've already shown a fragment of neural network definition, omitting the glue/noise. However, it's still very valuable IMHO that the QuickCheck tests exist for many small and big examples, including both examples in the README and many other mentioned in the README. And they are not smoke tests, but randomized checks of the fundamental formula you suggested

> dt <.> derivativeFromFun f ds = gradientFromFun dt f <.> ds

which the code was not designed to compute nor fine-tuned to pass. The tests pass as a happy coincidence, because the code is faithful to maths and they can be expressed as cleanly as above, because the API mimics maths.

Mikolaj (commented):
Thank you, that's a great improvement. Over to you. :)

README.md
```diff
@@ -1,19 +1,122 @@
...
+foo :: RealFloat a => (a,a,a) -> (a,a)
+foo (x,y,z) = 
+  let w = x * sin y 
+  in (atan2 z w, z * w)
+```
+TODO: show how the 2x3 Jacobian emerges from here
+
+# Forall shapes and sizes
+
+An additional feature of this library is a type system for tensor shape arithmetic.  Look at this beauty.
```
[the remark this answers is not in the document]
tomjaguarpaw: Isn't it true regardless of whether `orthotope` provides the heavy lifting for us?

README.md
```diff
@@ -1,19 +1,122 @@
...
+     let x : y : z : _ = vars variables 
+     w <- x * sin y 
+     returnLet $ (atan2 z w + z * w)  -- [note extra w, we *want* to require returnLet up front --awf] 
+   }
+```
+We hope to automate this transfomation soon, automatically generating `fooM` from `foo`, but for now it's a mechanical, if somewhat tedious, manual task.
+
+And then the gradient of `foo` is
+```Haskell
+grad_foo :: HasDelta r => Domain0 r -> (Domain0 r, r)
```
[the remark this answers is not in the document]
tomjaguarpaw: Yeah, we need to abolish `DualNumberVariables` and `Domain_` in favour of something like `AdaptArg` that I have in my version. That way we can preserve the tuple structure of the inputs and outputs. Still, my `AdaptArg` itself needs to be made more palatable, but it's the right direction to go.

README.md
Mikolaj: Right. It's finally published now.

README.md
```diff
@@ -1,19 +1,122 @@
...
+# HORDE-AD: Higher Order Reverse Derivatives Efficiently 
+
+This is an Automatic Differentiation library based on the paper _"Provably correct, asymptotically efficient, higher-order reverse-mode automatic differentiation"_ by Krawiec, Krishnaswami, Peyton Jones, Ellis, Fitzgibbon, and Eisenberg.
+
+This is an early prototype, both in terms of the engine performance and the API and toolbox for the library user. The user should be ready to add missing primitives, as well as obvious tools that should be predefined but aren't. One can already build with the library all basic neural network architectures, such as fully connected, recurrent, convolutional and residual networks. The library, due to its loose coupling of differentiation and data containers, can naturally handle exotic variants of such networks that wouldn't express well in a language of matrices or tensors. 
+
+Applications outside machine learning are very much in scope, given that the notion of a neural network is not hardwired into the formalism, but ad hoc redefined from basic blocks of general automatic differentiation whenever it's needed.
```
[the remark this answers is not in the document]
Mikolaj: Basic blocks is, on one hand, various differentiable functions that already come with the library: arithmetic, tensor ops, loss and activation functions, various kinds of data or weight normalization (#42), convolution-related ops, etc.. Then there are mechanisms for defining new differentiable functions (not just by combining the ones bundled with the library). Then there are operations that compute gradient, derivative, etc. Then various optimizers (currently only variants of gradient descent). The last two depend on the iffy mechanism (`DualNumberVariables`) for bundling parameters of various ranks and sizes into one value that can be passed along, taken derivative or gradient with respect to, mutated in gradient descent iterations, etc.

README.md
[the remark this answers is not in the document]
Mikolaj: Yes, that would be great (#48). I guess at rank 0 it would kill performance, but if so, we can keep both systems to be able to squeeze out the performance for that use case (a rare case, probably, but many toy benchmarks live there, including many of ours).

README.md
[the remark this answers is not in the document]
Mikolaj: BTW, would "basic building blocks" be more understandable?

README.md
```diff
@@ -1,19 +1,122 @@
...
+-- A function that goes from `R^3` to `R`. 
+-- [Arguments are tupled, could just as easily be curried. -awf]
+foo :: RealFloat a => (a,a,a) -> a
+foo (x,y,z) = 
+  let w = x * sin y 
+  in atan2 z w + z * w  -- [note extra w, we *want* to require returnLet up front --awf] 
+```
+
+In order to compute the gradient, we transform the function in the classic manner for forward-mode AD, to take `DualNumber` variables, 
+and the `DualMonad` which implements the system in the paper
```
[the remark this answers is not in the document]
Mikolaj: > But I'm still concerned about the types! Do we really lose the compile-time check that calls to foo are expected to take three variables?

Yes, we do. When you look at big shaped examples in https://github.com/Mikolaj/horde-ad/tree/master/src/HordeAd/Tool, the creative part of the code is nice and typed and composes in a typed way and at the end comes the glue code that explicitly defines tensor shapes that are already implicitly given in the tuple of your function arguments, squashes types, adds proxies and makes it all a mess. This could be handled by the plugin that is supposed to add the monad things. Or we could rework the types of the "basic building blocks" of the library (optimizers, generators of random weights, etc.) so that they don't expect a single tuple of vectors, but something more clever, e.g., related to @tomjaguarpaw's biapplicative `AdaptArg`.

Mikolaj:
This is now mostly done, with resulted discussed in #75 (please discuss some more). Thank you! Closing.

Issue #23, opened 2022-03-25
Unpack r in Delta datatypes via backpack, class or otherwise


Mikolaj:
This is low priority, because it's going to be a few (dozen) percent speedup (and a slight space saving and improved storage locality, which may be more important in real world situations). See https://hackage.haskell.org/package/unpacked-containers for a similar endeavour and https://twitter.com/kmett/status/982078739482738688 for a benchmarking anecdote.

I hope this now has a trivial solution and I was just not paying attention. Parameterized datatype not being a zero-cost abstraction is embarrassing. (Of course we could easily avoid the embarrassment by not implementing the datatype field upacking optimization in GHC, which few, if any, programming languages support.)

Mikolaj:
The fact that `DeltaS` requires a GADT makes this optimization problematic for the S rank (BTW, it would be interesting to benchmark S rank vs rank 1 and 2). The values contained in  `DeltaS` constructors are polymorphic (vs shape) and so have to be boxed and so can't be unpacked. However, we could instead keep in `DeltaS` untyped tensors or even the unwrapped monomorphic (once `r` is fixed) representations of shaped tensors and only retain the shapes in the `DeltaS` GADT. Whether one can unpack monomorphic values into GADT constructors is to be verified.

Mikolaj:
Since this ticket was created, we've had and have so many other performance bottlenecks that this one has become relatively unimportant. Closing.

PR #54, opened 2022-06-06
Convert Deltas to GADT

tomjaguarpaw:
@Mikolaj Are you open to converting the `Delta` types to a GADT? I suspect this is going to make a lot of things simpler in the long run. I haven't converted `Delta`s `2`, `X` and `S` here, but can do if you like the idea.

### DIFF
--- a/src/HordeAd/Internal/Delta.hs
+++ b/src/HordeAd/Internal/Delta.hs
@@ -725,54 +725,52 @@ derivativeFromDelta inlineDerivative0 inlineDerivative1 inlineDerivative2
                     inlineDerivativeX inlineDerivativeS
                     st deltaTopLevel
                     _ds@(params0Init, params1Init, params2Init, paramsXInit) =
-  let eval0 :: Domains r -> Delta0 r -> r
-      eval0 parameters@(params0, _, _, _) = \case
+  let eval :: Domains r -> Delta r t -> t
+      eval parameters@(params0, params1, _, _) = \case
         Zero0 -> 0
-        Scale0 k d -> k * eval0 parameters d
-        Add0 d e -> eval0 parameters d + eval0 parameters e
+        Scale0 k d -> k * eval parameters d
+        Add0 d e -> eval parameters d + eval parameters e
         Var0 (DeltaId i) -> params0 V.! i
 
-        SumElements0 vd _n -> HM.sumElements $ eval1 parameters vd
-        Index0 d ix _k -> eval1 parameters d V.! ix
+        SumElements0 vd _n -> HM.sumElements $ eval parameters vd
+        Index0 d ix _k -> eval parameters d V.! ix
 
-        Dot0 vr vd -> vr <.> eval1 parameters vd
+        Dot0 vr vd -> vr <.> eval parameters vd
 
         FromX0 d -> OT.unScalar $ evalX parameters d
         FromS0 d -> OS.unScalar $ evalS parameters d
 
         Outline0 codeOut primalArgs dualArgs ->
-          eval0 parameters $ inlineDerivative0 codeOut primalArgs dualArgs
-        Delay0 d -> eval0 parameters d
-      eval1 :: Domains r -> Delta1 r -> Vector r
-      eval1 parameters@(_, params1, _, _) = \case
+          eval parameters $ inlineDerivative0 codeOut primalArgs dualArgs
+        Delay0 d -> eval parameters d
         Zero1 -> 0
-        Scale1 k d -> k * eval1 parameters d
-        Add1 d e -> eval1 parameters d + eval1 parameters e
+        Scale1 k d -> k * eval parameters d
+        Add1 d e -> eval parameters d + eval parameters e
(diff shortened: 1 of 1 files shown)


goldfirere:
Why do you want to do this?

A GADT is a good idea when you might sometimes have a `Delta t1 t2` and pattern-matching would reveal whether it's a `Delta t t` or a `Delta t (Vector t)`. Yet I don't think that's ever the case here. Instead, we should always know, a priori, whether we have a scalar-like delta or a vector-like one. If that's right, then making this a GADT doesn't seem like the right direction.

Perhaps a middle ground is to use a data family:

```hs
data family Delta s t
data instance Delta r r = -- all the 0 constructors
data instance Delta r (Vector r) = -- all the 1 constructors
```

Except this doesn't quite work, because GHC is very nervous about the possibility that `r` might equal `Vector r`, in which case it would get confused between the two definitions. (`r` *can't* equal `Vector r`, because that requires an infinite type. But GHC still worries about the possibility. See Section 6.1 of https://richarde.dev/papers/2014/axioms/axioms.pdf.)

In any case, this doesn't quite smell like a case for GADTs to me.

tomjaguarpaw:
> Why do you want to do this?

It's rather hard to articulate exactly why but in [my version](https://github.com/Mikolaj/horde-ad/blob/d0d82174c618b1ee1d05851ac260fd06cbf8598d/src/HordeAd/Core/Again.hs) having a GADT rather than a collection of loosley-connected data types unifies and simplifies things tremendously. Ultimately I think the benefit derives from our `Delta` type being expressed a [free monad](https://hackage.haskell.org/package/free-5.1.4/docs/Control-Monad-Free.html) over an indexed functor. (The common-or-garden free monad is over a non-indexed functor, meaning it describes a "unityped" language. In the indexed version the indices are the types of expressions in the language, here scalar, vector, matrix etc..) In fact, because we have variables it's actually an indexed free monad with *scopes* (as per [bound](https://hackage.haskell.org/package/bound-2.0.5/docs/Bound.html#t:Scope)). 

> I don't think that's ever the case here

It's not the case _here_ because there isn't a GADT so we can't do what you suggest! But [it is the case in my version](https://github.com/Mikolaj/horde-ad/blob/d0d82174c618b1ee1d05851ac260fd06cbf8598d/src/HordeAd/Core/Again.hs#L166-L188). The benefit means we can avoid a type class constraint on `deltaLet`, whereas [Mikolaj's version picks one up](https://github.com/Mikolaj/horde-ad/blob/d0d82174c618b1ee1d05851ac260fd06cbf8598d/src/HordeAd/Core/DualNumber.hs#L42-L45). Since we're trying to reduce the number of mysterious concepts floating around in the API this seems rather important.

Further, I expect a number of awkward internal aspects of the library to dissolve on transition to GADT.

tomjaguarpaw:
Unifying the `eval0/1` functions is a good example of the unification I am talking about. See https://github.com/Mikolaj/horde-ad/pull/54/commits/fdd4c40368354ad62eb1d86c0b69f6aea5c2a793 and https://github.com/Mikolaj/horde-ad/pull/54/commits/368c296fd085b0c7e10a08cd025960abbaab5a9e.

Perhaps I can turn the question round: why should I _not_ want to do this? There's no performance impact (is there)? Conceptually `Delta0` and `Delta1` are the same thing, just at different types. Why not distinguish them (just) with a type index?

Mikolaj:
> Are you open to converting the Delta types to a GADT?

Potentially. I agree the simplification is enticing, but I'm a low-tech person, so I prefer repetitive dumb things the performance of which I understand to nifty things that are simple outside but complex inside. 

Inspired by your code, I started precisely your refactoring and complained in https://github.com/Mikolaj/horde-ad/issues/20#issuecomment-1133078472 I can't try out how big performance hit it would have, because our benchmarks are broken. Then I realized this refactoring may block #23 and complicate any future hacks that depend on simplicity (e.g., serialization or storing in memory in a more compressed form).

But if this gives us, e.g.,  something better than `DualNumberVariables` and `Domain_` without a plugin, I'm fine buying simplicity of plugin-free and TH-free code at the cost of complex in-memory representation of deltas and code more susceptible to GHC specialization bugs that silently degrade performance.

tomjaguarpaw:
Very well. Ultimately whatever API improvements we can get with GADTs we can also get using other tricks. I was mostly interested in simplifying the internals. Therefore I think I should keep GADTs for my version only.

Mikolaj:
> Therefore I think I should keep GADTs for my version only.

If it turns out the GADT formulation somehow synergizes with the other goodies you have, like `AdaptArg` for preserving types, I may yet become a happy customer. (Also, when our benchmarks are fixed and we can see clearly the cost.)

goldfirere:
I find @tomjaguarpaw's argument convincing in favor of using GADTs. Thank you.

As for performance: I wouldn't expect GADTs to have a negative impact on performance. A GADT constructor should have the same runtime effect as any other constructor. Maybe I'm missing something -- perf issues are often complex -- but I wouldn't worry about this one.

Issue #47, opened 2022-05-19
investigate huge floating point errors in (flaky) tests


sfindeisen:
Depending on the architecture and/or platform, we're sometimes getting test errors like:

```
FAIL (31.18s)
      test/common/TestMnistFCNN.hs:132:
      expected: 0.8991
       but got: 0.7424999999999999
      Use -p '/2 artificial 5 4 3 2 1 5 4 3 2 2393 1.0/' to rerun this test only.
```

This difference is **huuge**.

Let's try to investigate this and possibly also other similar test cases, understand what is going on, then fix the test case and/or the implementation.

sfindeisen:
More failures:

```
$ cabal test --enable-optimization shortTestForCI
[...]
FAIL (10.29s)
      test/common/TestMnistFCNN.hs:198:
      expected: 0.7756000000000001
       but got: 0.7882
      Use -p '/VV artificial 5 4 3 2 1 5 4 3 2 0 18 2393 1.0/' to rerun this test only.
[...]
FAIL (7.47s)
      test/common/TestMnistFCNN.hs:275:
      expected: 0.8085
       but got: 0.8114
      Use -p '/LL artificial 5 4 3 2 1 5 4 3 2 0 3 3 2393 1.0 0.5/' to rerun this test only.
[...]
FAIL (13.25s)
      test/common/TestMnistFCNN.hs:548:
      expected: 0.8246
       but got: 0.8202
      Use -p '/S artificial 5 4 3 2 1 5 4 3 2 6 2393 1.0 0.5/' to rerun this test only.
[...]
```

PR #57, opened 2022-06-15
implement epsilon-based floating point number comparisons in tests

### LINKED ISSUE #46: implement epsilon-based floating point number comparisons in tests
Depending on the architecture and/or platform, we're sometimes getting test errors like:

```
    black/white 250 209310 2.0e-2:                             FAIL (2.49s)
      test/common/TestMnistFCNN.hs:75:
      expected: 23.025850929940457
       but got: 23.02585092994046
```

This shows we need more relaxed floating point value equality checks in tests. Perhaps something like: `abs(x-y) < ε`, where `ε` should be configurable.

Within the scope of this ticket is:

1. implement the `ε` mechanism
2. make `ε` configurable
3. refactor all the flaky tests to use the new `ε`

Related issue: https://github.com/Mikolaj/horde-ad/issues/47

sfindeisen:
Implements `--eq-epsilon` flag for `cabal test` which sets the precision for floating point value comparison.

Example 1 (no flag):

```
$ cabal test shortTestForCI --test-options='-p "/randomLL2 140 0 3 5 54282 5.0e-2/"'
[...]
Short tests for CI
  MNIST RNN short tests
    randomLL2 140 0 3 5 54282 5.0e-2: OK (2.61s)
```

Example 2 (high precision):

```
$ cabal test shortTestForCI --test-options='-p "/randomLL2 140 0 3 5 54282 5.0e-2/" --eq-epsilon 0.0000000000000000001'
[...]
Short tests for CI
  MNIST RNN short tests
    randomLL2 140 0 3 5 54282 5.0e-2: FAIL (2.68s)
      test/common/TestCommon.hs:62:
      wrong result: 30.061871495725264 is expected to be a member of [30.061871495723956,30.06187089990927]
```

Fixes #46


test/common/TestCommon.hs
```diff
@@ -9,15 +10,82 @@ import Prelude
...
 import HordeAd.Core.DualClass (Dual)
 
+-- Default value for eqEpsilonRef
+eqEpsilonDefault :: Double
+eqEpsilonDefault = 1e-6
+
+-- Ugly global epsilon used to compare floating point values.
+eqEpsilonRef :: IORef Double
+{-# NOINLINE eqEpsilonRef #-}
+eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
```
Mikolaj, line 32: I'm proud of you.

sfindeisen:
Wait a sec with the review, I will tweak it a bit more.

test/common/TestCommon.hs
sfindeisen: It took me hours. It's so tricky! Did you know, for example, that this version:

```haskell
epsilonRef :: Fractional a => IORef a
{-# NOINLINE epsilonRef #-}
epsilonRef = unsafePerformIO $ newIORef 3.1415
```

doesn't work?! Even if I explicitly write `:: Double` everywhere I use it. Why is that?!

sfindeisen:
We have quite a lot of `@?=` in our testing code. What would be better: to hide+redefine this operator or to replace its occurences with the new `assertClose`?

This PR fixes 9 out of 13 `shortTestForCI` failures on my machine (the remaining 4 are subject to #47) and I do not use the other test suites so we can also leave it as-is.

test/common/TestCommon.hs
Mikolaj: I think you try to construct a value of an existential type `Exist a. Fractional a => IORef a`. However, to construct a value of an existential type you need to provide that type, not only a value. And you don't do that.

I'm unclear about the details of existential vs universal types in Haskell, e.g., if it's about positive occurrences of a type variable, or strictly positive and what happens if a variable has mixed variance. However, in this example it's simple. The occurrence of `a` is not under any number of arrows, so the whole thing is existential, not universal. If it was existential, it could be abused: you could construct a reference by initializing it with a Float, but later update it with a Double, overwriting a nearby memory cell (assuming Floats are smaller than Doubles).

Mikolaj:
> We have quite a lot of @?= in our testing code. What would be better: to hide+redefine this operator or to replace its occurences with the new assertClose?

How about `@?~` and `@~?`? I've searched hoogle to make sure it's not popular in another context and I've found

https://hackage.haskell.org/package/HUnit-approx-1.1.1.1/docs/Test-HUnit-Approx.html#v:-64--63--126-

Mikolaj:
And that package is by our own @goldfirere, no less!

test/common/TestCommon.hs
tomjaguarpaw: GHC implements type classes as dictionary passing, so

```haskell
epsilonRef :: Fractional a => IORef a
epsilonRef = unsafePerformIO $ newIORef 3.1415
```

is rather similar to

```haskell
epsilonRef :: Fractional a -> IORef a
epsilonRef fractionalDict = unsafePerformIO $ newIORef (fromRational fractionalDict 3.1415)
```

and it seems rather likely that this makes a fresh `IORef` containing `3.1415` at each use site. Would that explain what you see?

sfindeisen:
> And that package is by our own @goldfirere, no less!

If this is the case, shall we try to use it instead of those custom-made `assertClose*` functions which I have just implemented?

test/common/TestCommon.hs
Mikolaj: > `Fractional a -> IORef a`

In that case, type variable `a` has an occurrence on the left had side of an arrow so, yes, this could probably be a universal type if interpreted that way. Similarly, if this function took an argument to initialize the ref with: `a -> IORef a`. And if these are lambdas, the body of a lambda is never evaluated, so the ref is not created. But I guess

```
epsilonRef :: IORef a
epsilonRef = unsafePerformIO $ newIORef undefined
```

would be firmly in the existential territory and no lambda, so perhaps my explanation only applied to that one. Not tested.

Mikolaj:
> > And that package is by our own @goldfirere, no less!
> 
> If this is the case, shall we try to use it instead of those custom-made `assertClose*` functions which I have just implemented?

Certainly, though I doubt it's compatible with tasty. In any case, we may learn something that way.

test/common/TestCommon.hs
sfindeisen: @tomjaguarpaw I can't comment on GHC's type classes implementation but yes, what I see is like a fresh `IORef` each time.

Mikolaj:
> If the chance of success is so low, shall we make it a separate issue? Let's merge this fix first, and then we can continue with speculative experiments.

Of course. Will review soon.

Mikolaj:
But what about

> > We have quite a lot of @?= in our testing code. What would be better: to hide+redefine this operator or to replace its occurences with the new assertClose?
> 
> How about `@?~` and `@~?`?

sfindeisen:
> > How about `@?~` and `@~?`?

You mean, defining our own?

Yes sure, can do. If it improves legibility... ;)

Mikolaj:
> Yes sure, can do. If it improves legibility... ;)

I think it does and also wouldn't it be a drop-in replacement for "quite a lot of @?= in our testing code". The diff would be smaller and people accustomed to Tasty.Hunit would have an easier time reading this.

test/common/TestCommon.hs
sfindeisen: Here's a complete example:

```haskell
import Data.IORef
import System.IO.Unsafe

epsilonRef :: Fractional a => IORef a
{-# NOINLINE epsilonRef #-}
epsilonRef = unsafePerformIO $ newIORef 3.1415

setEpsilon :: Fractional a => a -> IO ()
setEpsilon = atomicWriteIORef epsilonRef

main :: IO ()
main = do
  () <- (setEpsilon :: (Double -> IO ())) (0.2 :: Double)  -- doesn't work!
  epsilon <- (readIORef (epsilonRef :: IORef Double))
  print $ "epsilon: " ++ show epsilon
```

On my machine this prints:

```
"epsilon: 3.1415"
```

Is this what you'd expect?

sfindeisen:
Look how weird we have it now on the command line: `--eq-epsilon "1 % 10000000000000000000"`. There must be a way to fix it, let me check...

sfindeisen:
> Look how weird we have it now on the command line: `--eq-epsilon "1 % 10000000000000000000"`. There must be a way to fix it, let me check...

Fixed. But internally, `ε` is a `Rational` now. Then the polymorphic type (`a`) used in comparisons is `Fractional a`, and we can say `< fromRational ε`...

Mikolaj:
> I implemented `@?~` infix operator (priority: `1`; is this correct?

I think, let's emulate the priority of

https://hackage.haskell.org/package/tasty-hunit-0.10.0.3/docs/Test-Tasty-HUnit.html#v:-64--61--63-

which is 1.

Also, they seem to use Eq to recursively descend into complex structures, so perhaps let's define our own class, with a few obvious instances? I don't know how Eq is defined, but perhaps it uses Traversable or something like that, so maybe let's generalize your handling of lists to Traversable (or whatever it is)?

sfindeisen:
> I assume this is ready for review?

Yes please.

> I'm surprised: HUnit-approx could be reused, after all, despite tasty not using [HUnit](https://hackage.haskell.org/package/HUnit)? They are compatible enough regardless? That's great news. Is it because their internal types are not abstract and they agree? Was your initial implementation of `assertApproxEqual` approximately equal to that from HUnit-approx?

4x yes

goldfirere:
Oh, wow. I didn't think I'd be hearing from `HUnit-approx` again! :) Glad it was useful.

Mikolaj:
> Oh, wow. I didn't think I'd be hearing from `HUnit-approx` again! :) Glad it was useful.

We'll still make it big. :D

sfindeisen:
Thanks for the review @Mikolaj . What about the open discussions, esp. the one about `Data.Vector.Generic`? Will you have any comments there?

Mikolaj:
> Thanks for the review @Mikolaj . What about the open discussions, esp. the one about `Data.Vector.Generic`? Will you have any comments there?

Huh, github is not showing me anything. Could you give me links?

sfindeisen:
> Huh, github is not showing me anything. Could you give me links?

If you click on `Files changed` tab then you'll see there `Pending` discussions.

Mikolaj:
I can't see anything. The "Conversations" drop down menu shows an irrelevant discussion. Do you mean a different PR? I tried in a different browser, not logged in to github, but still nothing.

sfindeisen (commented):
^^ @Mikolaj

test/common/TestCommonEqEpsilon.hs
```diff
@@ -0,0 +1,119 @@
...
+setEpsilonEq (EqEpsilon x) = atomicWriteIORef eqEpsilonRef x
+
+-- | Asserts that the specified actual floating point value is close to the expected value.
+-- The output message will contain the prefix, the expected value, and the
+-- actual value.
+--
+-- If the prefix is the empty string (i.e., @\"\"@), then the prefix is omitted
+-- and only the expected and actual values are output.
+assertClose :: forall a. (Fractional a, Ord a, Show a, HasCallStack)
+            => String -- ^ The message prefix
```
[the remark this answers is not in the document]
Mikolaj: Yes, I'd do that. If we suddenly need them, we can re-add.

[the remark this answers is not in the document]
Mikolaj: Or if tasty devs request it as a condition for including the new operations in the tasty package. Perhaps ask them now?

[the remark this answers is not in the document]
sfindeisen: Asked them [here](https://github.com/UnkindPartition/tasty/issues/337#issuecomment-1172714850).

Mikolaj:
Thank you for the experiment. I turns out the instance is missing, except for boxed vectors. I wonder what deep reason is there for this and if we could add an orphan instance similar to the one for boxed vectors: https://hackage.haskell.org/package/vector-0.13.0.0/docs/src/Data.Vector.html#line-461

test/common/TestSingleGradient.hs
```diff
@@ -427,9 +428,11 @@ readmeTests :: TestTree
 readmeTests = testGroup "Simple tests for README"
   [ testCase " Float (1.1, 2.2, 3.3)"
     $ atanReadmeDReverse (V.fromList [1.1 :: Float, 2.2, 3.3])
+      -- TODO: change this to @?~
```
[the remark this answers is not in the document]
Mikolaj: The latter error is caused by boxed vectors having a Traversable instance and so providing an alternative instance of `AssertClose`. But we don't use boxed vectors, so this instance via Traversable is useless for us.

Mikolaj:
Or, alternatively, you could provide the AssertClose instance only for Storable vectors, which is what we use.

sfindeisen:
First of all, our `vector` package version seems to be `vector-0.12.3.1`. But I can't see this version number anywhere: where is it coming from? I can see a lot of crap like:

```
$ find . -type f -name '*cabal*' -print0 | xargs -0 grep -i vector 
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:/* package vector-0.12.3.1 */
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#ifndef VERSION_vector
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#define VERSION_vector "0.12.3.1"
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#endif /* VERSION_vector */
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#ifndef MIN_VERSION_vector
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#define MIN_VERSION_vector(major1,major2,minor) (\
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/build/autogen/cabal_macros.h:#endif /* MIN_VERSION_vector */
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:/* package vector-0.12.3.1 */
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#ifndef VERSION_vector
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#define VERSION_vector "0.12.3.1"
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#endif /* VERSION_vector */
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#ifndef MIN_VERSION_vector
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#define MIN_VERSION_vector(major1,major2,minor) (\
./dist-newstyle/build/x86_64-linux/ghc-9.2.2/horde-ad-0.1.0.0/l/testLibrary/build/testLibrary/autogen/cabal_macros.h:#endif /* MIN_VERSION_vector */
./horde-ad.cabal:                      HordeAd.Core.PairOfVectors
./horde-ad.cabal:                      HordeAd.Tool.MnistFcnnVector
./horde-ad.cabal:      , vector
./horde-ad.cabal:      , vector
./horde-ad.cabal:      , vector
```

The latest version, as you pointed out, is `vector-0.13.0.0`.

Mikolaj:
> But I can't see this version number anywhere: where is it coming from?

Cabal solved the constraints and that's (close to) the newest version permissible given what other packages require. You can see some of the process with `cabal build -v2`, though there may be better flags for that.

sfindeisen:
> Or, alternatively, you could provide the AssertClose instance only for Storable vectors, which is what we use.

That's how I fixed it. Thanks. OK to merge?

tomjaguarpaw:
I haven't really been following, so nothing to add from me!

Mikolaj:
Thank you Staszek. Thank you Tom. The branch seems rebased so, Staszek, please feel free to merge (perhaps with the squash method).

PR #59, opened 2022-07-05
Voice recognition (fixes #58)

### LINKED ISSUE #58: Code an example that recognizes a particular person's speech
A possible formulation: given a few minutes of a person X speech, the network should be able to quickly determine if any short audio recording contains speech by X or not. Listening to the sample of X speech is permitted to take a lot of resources (real training), but the subsequent classification of many audio fragments should be cheap (no training any more).

The network could initially train on speech samples of many persons that don't include X speech, unless it doesn't make it more accurate nor faster in learning a new speech pattern nor faster in classifying. If that initial step is beneficial, the important question is what data we need for training and if unlabelled data suffices, how we process the data, where we get labelled data for testing.

To determine: loss function, how to get data for training (in addition to the X speech), the architecture of the neural network.

Mikolaj:
Implements #58. An attempt to recognize to which person a voice belongs in a given window of a sound file. Uses RNN. 

The tests to run are `cabal test extremelyLongTest --enable-optimization -f test_seq --test-options='-p "Speech RNN"''`. Example data files not yet included.

### DIFF
--- a/horde-ad.cabal
+++ b/horde-ad.cabal
@@ -204,6 +204,8 @@ library testLibrary
     -- Other library packages from which modules are imported.
     build-depends:
         base
+      , cereal
+      , bytestring
       , deepseq
       , HUnit-approx
       , hmatrix
--- a/horde-ad.cabal
+++ b/horde-ad.cabal
@@ -196,6 +196,7 @@ library testLibrary
     exposed-modules:  TestMnistCNN
                       TestMnistFCNN
                       TestMnistRNN
+                      TestSpeechRNN
                       TestSimpleDescent
                       TestSingleGradient
   }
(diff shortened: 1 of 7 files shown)


Mikolaj:
I'm afraid, it's very outdated and the new tools it should use instead of the old ones are not ready.

blackhole64:
Which tools? Is there a PR/issue for those?

Mikolaj:
To be frank, it's not yet clear what those tools should be. Benchmarks are going to show if more work is needed, but benchmarks first need to be ported from the old API and also new ones created. The new API is not finished yet, though. Once it's finished in its current form, ideally we'd add a shaped, not only ranked version, but this may be too hard to do, at least initially. Once we decide and implement that, benchmarking is the next step.

PR #67, opened 2022-09-17
implement equality testing with explicit error margin

### LINKED ISSUE #65: Change of spec for assertClose: it needs to take eps as an argument
The client has completely changed the spec for testing up to eps (#64; or rather the management didn't ask him in time and now finds excuses). Now `assertClose` should be called `assertEqualUpToEps`, takes the eps as the first argument, should work on many types, just as `@?~` did previously, which now should differ only in that it uses the default eps from the `IORef`. The accepted types should include tuples (up to 10 elements, say) of any types accepted previuously.

@sfindeisen: you have dibs on this as the author of the subsystem, but please feel free to skip the privilege and instead take something harder, e.g., the generic library thing from #64 [edit: started in #66].

sfindeisen:
Implements fp-equality testing with the error margin explicitly passed as a parameter from the caller. For example:

```haskell
testFoo :: Assertion
testFoo =
  assertEqualUpToEpsilon (1e-10 :: Double)
    (rev foo (1.1, 2.2, 3.3) :: (Double,Double,Double))
    (2.4396285219055063, -1.953374825727421, 0.9654825811012627)
```

This enables us to specify different error margins for different comparisons. If you want to keep it simple, you can still rely on the default error margin specified on the command line using `--eq-epsilon`. In that case, you can use the binary `@?~` operator:

```haskell
res @?~ ([3.5e-44,3.5e-44],5.0 :: Float)
```

This also improves error messages in accordance with #70 .

Fixes #65 #70

### DIFF
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -192,5 +192,5 @@ instance (VS.Storable a) => Linearizable (OT.Array a) a where
 instance (VS.Storable a, OS.Shape sh) => Linearizable (OS.Array sh a) a where
   linearize = OS.toList
 
-instance {-# OVERLAPPABLE #-} (Foldable t) => Linearizable (t a) a where
-  linearize = foldr (:) []
+instance (VS.Storable a, HM.Element a) => Linearizable (HM.Matrix a) a where
+  linearize = HM.toList . HM.flatten
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -192,5 +192,5 @@ instance (VS.Storable a) => Linearizable (OT.Array a) a where
 instance (VS.Storable a, OS.Shape sh) => Linearizable (OS.Array sh a) a where
   linearize = OS.toList
 
-instance (VS.Storable a, HM.Element a) => Linearizable (HM.Matrix a) a where
-  linearize = HM.toList . HM.flatten
+instance {-# OVERLAPPABLE #-} (Foldable t) => Linearizable (t a) a where
+  linearize = foldr (:) []
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -10,6 +10,7 @@ import Prelude
 
 import qualified Data.Array.DynamicS as OT
 import qualified Data.Array.ShapedS as OS
+import qualified Data.Foldable
 import qualified Data.Vector.Generic as V
 import qualified Data.Vector.Storable as VS
 import qualified Numeric.LinearAlgebra as HM
@@ -199,4 +200,4 @@ instance (VS.Storable a, HM.Element a) => Linearizable (HM.Matrix a) a where
   linearize = HM.toList . HM.flatten
 
 instance {-# OVERLAPPABLE #-} (Foldable t) => Linearizable (t a) a where
-  linearize = foldr (:) []
+  linearize = Data.Foldable.toList
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -166,7 +166,7 @@ class HasShape a where
   shapeL :: a -> ShapeL
 
 instance (VS.Storable a) => HasShape (VS.Vector a) where
-  shapeL vec = [VS.length vec]
+  shapeL = (: []) . VS.length
 
 instance HasShape (OT.Array a) where
   shapeL = OT.shapeL
@@ -175,7 +175,7 @@ instance HasShape (HM.Matrix a) where
   shapeL matrix = [HM.rows matrix, HM.cols matrix]
 
 instance {-# OVERLAPPABLE #-} (Foldable t) => HasShape (t a) where
-  shapeL = (.) (flip (:) []) length
+  shapeL = (: []) . length
 
 ----------------------------------------------------------------------------
 -- Things that can be linearized, i.e. converted to a list
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -1,13 +1,17 @@
-{-# LANGUAGE DataKinds, RankNTypes, TypeFamilies #-}
+{-# LANGUAGE DataKinds, FlexibleInstances, FunctionalDependencies, MultiParamTypeClasses, RankNTypes, TypeFamilies #-}
 module TestCommon (listsToParameters,
                    cmpTwo, cmpTwoSimple,
-                   qcPropDom, quickCheckTest0, fquad, quad
+                   qcPropDom, quickCheckTest0, fquad, quad,
+                   HasShape, shapeL,
+                   Linearizable, linearize
                   ) where
 
 import Prelude
 
 import qualified Data.Array.DynamicS as OT
+import qualified Data.Array.ShapedS as OS
 import qualified Data.Vector.Generic as V
+import qualified Data.Vector.Storable as VS
 import qualified Numeric.LinearAlgebra as HM
 import           Test.Tasty
 import           Test.Tasty.QuickCheck
@@ -149,3 +153,38 @@ qcTestRanges txt f fArgDom dsRange perturbationRange dtRange =
   forAll (choose perturbationRange) $ \perturbationRaw ->
   forAll (choose dtRange) $ \dt ->
   ioProperty $ qcPropFArg f fArgDom xyz dsRaw perturbationRaw dt
+
+----------------------------------------------------------------------------
+-- Things that have shape
+----------------------------------------------------------------------------
+
+-- | Shape of the thing, typically a list of the sizes of its dimensions.
+type ShapeL = [Int]
+
+class HasShape a where
+  shapeL :: a -> ShapeL
+
+instance HasShape (OT.Array a) where
+  shapeL = OT.shapeL
+
+instance (OS.Shape sh) => HasShape (OS.Array sh a) where
+  shapeL = OS.shapeL
+
+instance HasShape (HM.Matrix a) where
+  shapeL matrix = [HM.rows matrix, HM.cols matrix]
+
+----------------------------------------------------------------------------
+-- Things that can be linearized, i.e. converted to a list
+----------------------------------------------------------------------------
+
+class Linearizable a b | a -> b where
+  linearize :: a -> [b]
+
+instance (VS.Storable a) => Linearizable (OT.Array a) a where
+  linearize = OT.toList
+
+instance (VS.Storable a, OS.Shape sh) => Linearizable (OS.Array sh a) a where
+  linearize = OS.toList
+
+instance (VS.Storable a, HM.Element a) => Linearizable (HM.Matrix a) a where
+  linearize = HM.toList . HM.flatten
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -1,5 +1,5 @@
 {-# LANGUAGE DataKinds, FlexibleInstances, FunctionalDependencies, RankNTypes, TypeFamilies #-}
-module TestCommon (listsToParameters,
+module TestCommon (lowercase, listsToParameters,
                    cmpTwo, cmpTwoSimple,
                    qcPropDom, quickCheckTest0, fquad, quad,
                    HasShape (shapeL),
@@ -10,6 +10,7 @@ import Prelude
 
 import qualified Data.Array.DynamicS as OT
 import qualified Data.Array.ShapedS as OS
+import qualified Data.Char
 import qualified Data.Foldable
 import qualified Data.Vector.Generic as V
 import qualified Data.Vector.Storable as VS
@@ -19,6 +20,9 @@ import           Test.Tasty.QuickCheck
 
 import HordeAd hiding (sumElementsVectorOfDual)
 import HordeAd.Core.DualClass (Dual)
+
+lowercase :: String -> String
+lowercase = map Data.Char.toLower
 
 -- Checks if 2 numbers are close enough.
 close1 :: forall r. (Ord r, Fractional r)
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -171,9 +171,6 @@ instance (VS.Storable a) => HasShape (VS.Vector a) where
 instance HasShape (OT.Array a) where
   shapeL = OT.shapeL
 
-instance (OS.Shape sh) => HasShape (OS.Array sh a) where
-  shapeL = OS.shapeL
-
 instance HasShape (HM.Matrix a) where
   shapeL matrix = [HM.rows matrix, HM.cols matrix]
 
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -1,4 +1,4 @@
-{-# LANGUAGE DataKinds, FlexibleInstances, FunctionalDependencies, MultiParamTypeClasses, RankNTypes, TypeFamilies #-}
+{-# LANGUAGE DataKinds, FlexibleInstances, FunctionalDependencies, RankNTypes, TypeFamilies #-}
 module TestCommon (listsToParameters,
                    cmpTwo, cmpTwoSimple,
                    qcPropDom, quickCheckTest0, fquad, quad,
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -2,8 +2,8 @@
 module TestCommon (listsToParameters,
                    cmpTwo, cmpTwoSimple,
                    qcPropDom, quickCheckTest0, fquad, quad,
-                   HasShape, shapeL,
-                   Linearizable, linearize
+                   HasShape (shapeL),
+                   Linearizable (linearize)
                   ) where
 
 import Prelude
--- a/test/common/TestCommon.hs
+++ b/test/common/TestCommon.hs
@@ -176,6 +176,9 @@ instance (OS.Shape sh) => HasShape (OS.Array sh a) where
 instance HasShape (HM.Matrix a) where
   shapeL matrix = [HM.rows matrix, HM.cols matrix]
 
+instance {-# OVERLAPPABLE #-} (Foldable t) => HasShape (t a) where
+  shapeL = (.) (flip (:) []) length
+
 ----------------------------------------------------------------------------
 -- Things that can be linearized, i.e. converted to a list
 ----------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


Mikolaj:
When you reach the point of writing `AssertClose` instances, to be used in `assertEqualUpToEps`, for `Vector` or `MonoTraversable` (it's the same thing), let me know, I'd share what I know.

sfindeisen:
> When you reach the point of writing `AssertClose` instances, to be used in `assertEqualUpToEps`, for `Vector` or `MonoTraversable` (it's the same thing), let me know, I'd share what I know.

Yes please. I was looking into that, and it looks we cannot simply use `Traversable` because it doesn't work with pairs (they way we do):

```haskell
ghci> import Data.Foldable 
ghci> traverse_ (\x -> print x) ("abc", "def")
"def"
```

This is fine:

```haskell
ghci> traverse_ (\x -> print x) ["abc", "def"]
"abc"
"def"
```

and could be used.

Mikolaj:
Yes, unfortunately, we probably can't do better than have 10 instances of the type class for tuples up to 10. A pity. We should have something like Traversable for tuples, but it may be too hard to type, because it's not tuples of the same type. But even something using Dynamic would help.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,6 +37,134 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
...
+             -> OS.Array sh a         -- ^ The expected value
+             -> OS.Array sh a         -- ^ The actual value
+             -> Assertion
+assert_shape make_assert expected actual =
+  if shapeE == shapeA then
+    assert_list make_assert (OS.toList expected) (OS.toList actual)
+  else
+    assertFailure $ "Expected shape: " ++ show shapeE ++ ", but got: " ++ show shapeA
+  where
+    shapeE = OS.shapeL expected
```
Mikolaj, line 80: This will be needed for untyped tensors (`OT.Array`), for which we need an instance as well. But for shaped tensors, the types system guarantees the shapes agree.

BTW, what you wrote shows me that using a MonoTraversable instance is not enough for untyped tensors, because we don't want equate tensors that have different shapes but, by chance, the same list of elements. That's a pity.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,6 +37,134 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
...
+    assertEqualUpToEpsilon eqEpsilon e3 a3 >>
+    assertEqualUpToEpsilon eqEpsilon e4 a4
+
+instance {-# OVERLAPPABLE #-} (Traversable t, AssertEqualUpToEpsilon z a) => AssertEqualUpToEpsilon z (t a) where
+  assertEqualUpToEpsilon :: z -> t a -> t a -> Assertion
+  assertEqualUpToEpsilon eqEpsilon expected actual =
+    assert_list (assertEqualUpToEpsilon eqEpsilon) (asList expected) (asList actual)
+
+instance {-# OVERLAPPABLE #-} (VS.Storable a, AssertEqualUpToEpsilon z a) => AssertEqualUpToEpsilon z (VS.Vector a) where
+  assertEqualUpToEpsilon :: z -> VS.Vector a -> VS.Vector a -> Assertion
```
Mikolaj, line 156: We'd also need `Numeric.LinearAlgebra.Matrix`.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -70,26 +198,15 @@ assertCloseElem preface expected actual = do
...
-    go_assert (_:_) [] = assertFailure msgneq
-    go_assert (head_exp:tail_exp) (head_act:tail_act) =
-      (@?~) head_act head_exp >> go_assert tail_exp tail_act
+assertCloseList = assert_list (flip (@?~))
 
--- | Foldable to list.
-asList :: Foldable t => t a -> [a]
-asList = foldr (:) []
+----------------------------------------------------------------------------
+-- AssertClose class together with (@?~) operator
```
Mikolaj, line 208: Why not define `@?~` in terms of `assertEqualUpToEpsilon`?

test/common/TestSingleGradient.hs
```diff
@@ -423,8 +423,8 @@ foo (x,y,z) =
 
 testFoo :: Assertion
 testFoo =
-  assertEqualUpToEps (1e-10 :: Double)
-    (rev foo (1.1, 2.2, 3.3))
+  assertEqualUpToEpsilon (1e-10 :: Double)
+    (rev foo (1.1, 2.2, 3.3) :: (Double,Double,Double))
```
Mikolaj, line 427: It would be great to get rid of the extra type annotation. Perhaps define the class for `Double` and `Float` instead of for `a`? Perhaps then you can get rid of `OVERLAPPABLE` as well?

Mikolaj (commented):
Well done. I was worried the shaped part is going to get complex, but you made it simple (even after untyped tensors get added).

test/common/TestCommonEqEpsilon.hs
sfindeisen: I am trying things similar to the snippet below. But apparently I am having problems convincing Haskell to unify the type variables `z` and `z0`:

```haskell
instance {-# OVERLAPPABLE #-} forall a z . (Fractional z, AssertEqualUpToEpsilon z a) => AssertClose a where
  (@?~) :: forall {z} . a -> a -> Assertion
  (@?~) actual expected = do
    eqEpsilon <- readIORef eqEpsilonRef :: Rational
    assertEqualUpToEpsilon ((fromRational (eqEpsilon :: Rational)) :: z) (expected :: a) (actual :: a)
```

This is pretty verbose with respect to types, as I was trying to disambiguate it as much as I could. But still I am getting this error:

```
test/common/TestCommonEqEpsilon.hs:223:31: error:
    • Could not deduce (Fractional z0)
      from the context: (Fractional z, AssertEqualUpToEpsilon z a)
        bound by an instance declaration:
                   forall a z.
                   (Fractional z, AssertEqualUpToEpsilon z a) =>
                   AssertClose a
        at test/common/TestCommonEqEpsilon.hs:223:31-102
      The type variable ‘z0’ is ambiguous
    • In the ambiguity check for an instance declaration
      To defer the ambiguity check to use sites, enable AllowAmbiguousTypes
      In the instance declaration for ‘AssertClose a’
    |
223 | instance {-# OVERLAPPABLE #-} forall a z . (Fractional z, AssertEqualUpToEpsilon z a) => AssertClose a where
    |
```

What am I doing wrong here?

Mikolaj: First, why not remove the `AssertClose` class altogether and use only `AssertEqualUpToEpsilon`? Second, this is tough in general, but at this point I'd try adding the `Fractional` constraint to `(@?~)`. The variable `z` in this function has no relation to the outer `z` (because it's in an inner binding) and so doesn't have the constraint, so GHC correctly complains.

Mikolaj: Or perhaps remove the inner `forall {z}`, but I guess you tried that already. However, the error message may then be more to the point.

test/common/TestCommonEqEpsilon.hs
sfindeisen: Done

test/common/TestSingleGradient.hs
sfindeisen: Hm I did just that, but it doesn't help. :/

Mikolaj: And what's the error message?

sfindeisen:
```
test/common/TestSingleGradient.hs:428:6: error:
    • Couldn't match type: Dual
                             'ADModeGradient
                             (mono-traversable-1.0.15.3:Data.MonoTraversable.Element a1)
                     with: HordeAd.Internal.Delta.Delta0
                             (mono-traversable-1.0.15.3:Data.MonoTraversable.Element a1)
        arising from a use of ‘rev’
      The type variable ‘a1’ is ambiguous
    • In the second argument of ‘assertEqualUpToEpsilon’, namely
        ‘(rev foo (1.1, 2.2, 3.3))’
      In the expression:
        assertEqualUpToEpsilon
          (1e-10 :: Double) (rev foo (1.1, 2.2, 3.3))
          (2.4396285219055063, - 1.953374825727421, 0.9654825811012627)
      In an equation for ‘testFoo’:
          testFoo
            = assertEqualUpToEpsilon
                (1e-10 :: Double) (rev foo (1.1, 2.2, 3.3))
                (2.4396285219055063, - 1.953374825727421, 0.9654825811012627)
    |
428 |     (rev foo (1.1, 2.2, 3.3))
    |
```

Mikolaj: Thank you. That looks hard. After you merge, I will try to remember how I use the `MonoTraversable` and experiment to see if I can help the type family `Dual` resolve.

test/common/TestCommonEqEpsilon.hs
sfindeisen: How about now? With my new `HasShape` class?...

sfindeisen: Shape handling has been unified.

sfindeisen:
> Yes, unfortunately, we probably can't do better than have 10 instances of the type class for tuples up to 10. A pity. We should have something like Traversable for tuples, but it may be too hard to type, because it's not tuples of the same type. But even something using Dynamic would help.

One could perhaps squeeze those tuples into `HasShape` and `Linearizable`, if that makes sense...

Mikolaj:
> One could perhaps squeeze those tuples into `HasShape` and `Linearizable`, if that makes sense...

`Linearizable` indeed should be more succinct than `AssertEqualUpToEpsilon`.

test/common/TestCommonEqEpsilon.hs
Mikolaj: `HasShape` tersely covers the case of `Matrix` and `OT.Array` and gives a unified error message for vectors as well (which means, you can remove the instance for `VS.Vector`, right?) However, checking shapes of `OS.Array` is still spurious, because `OS.shapeL` reads the list from the type and the two type are equal there.

Mikolaj: Even if somebody did `unsafeCoerce` and the arrays underneath have different shapes.

sfindeisen:
> `Linearizable` indeed should be more succinct than `AssertEqualUpToEpsilon`.

... well but this will not work of course, because tuples can be heterogenous whereas `Linearizable` gives `[a]`. Sorry for the wrong suggestion.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,23 +39,136 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
...
-  assertBool (msg eqEpsilon) (abs (expected-actual) < fromRational eqEpsilon)
+assert_close_eps :: (Num a, Ord a, Show a, HasCallStack)
+                   => String -- ^ The message prefix
+                   -> String -- ^ The message suffix
+                   -> a      -- ^ The error margin
+                   -> a      -- ^ The expected value
+                   -> a      -- ^ The actual value
+                   -> Assertion
+assert_close_eps preface epilogue eqEpsilon expected actual = do
+  assertBool (msg eqEpsilon) (abs (expected-actual) < eqEpsilon)
```
Mikolaj, line 101: Just double-checking: would people expect `<` here or `<=`? With floating point it shouldn't matter much, but it may sometimes.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,23 +39,136 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
...
+                                -> a      -- ^ The expected value
+                                -> a      -- ^ The actual value
+                                -> Assertion
+
+  assertEqualUpToEpsilon :: z -- ^ The error margin (i.e., the epsilon)
+                         -> a -- ^ The expected value
+                         -> a -- ^ The actual value
+                         -> Assertion
+  assertEqualUpToEpsilon error_margin expected actual =
+    assertEqualUpToEpsilonWithMsg ("Expected: " ++ (show expected) ++ "\n but got: " ++ (show actual)) error_margin expected actual
```
Mikolaj, line 124: Good idea to have the two methods here.

test/common/TestCommon.hs
```diff
@@ -149,3 +153,50 @@ qcTestRanges txt f fArgDom dsRange perturbationRange dtRange =
...
+  shapeL :: a -> ShapeL
+
+instance (VS.Storable a) => HasShape (VS.Vector a) where
+  shapeL vec = [VS.length vec]
+
+instance HasShape (OT.Array a) where
+  shapeL = OT.shapeL
+
+instance (OS.Shape sh) => HasShape (OS.Array sh a) where
+  shapeL = OS.shapeL
```
Mikolaj, line 174: Do we use this instance anywhere?

Mikolaj (approved):
Thanks a lot. No more embarrassing copy-pastes needed for each new differentiation example. Please merge when you are ready.

test/common/TestCommon.hs
```diff
@@ -149,3 +153,50 @@ qcTestRanges txt f fArgDom dsRange perturbationRange dtRange =
...
+
+instance HasShape (HM.Matrix a) where
+  shapeL matrix = [HM.rows matrix, HM.cols matrix]
+
+instance {-# OVERLAPPABLE #-} (Foldable t) => HasShape (t a) where
+  shapeL = (.) (flip (:) []) length
```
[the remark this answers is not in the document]
sfindeisen: This is an `OVERLAPPABLE` instance to be used with `Foldable` as a fallback... in order to avoid typing instances for things like `(Data.Vector.Vector (Data.Vector.Storable.Vector Float))`. The one for the matrix is right above it and is stricly more specific so it will get selected, correct?

test/common/TestCommonEqEpsilon.hs
sfindeisen: Well it could matter e.g. if `eqEpsilon` is `0`. (But on the other hand, we allow it to be negative, i.e. there is no check.) Well spotted, fixed.

test/common/TestCommon.hs
sfindeisen: Nope. Fixed.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,23 +39,136 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
 setEpsilonEq :: EqEpsilon -> IO ()
 setEpsilonEq (EqEpsilon x) = atomicWriteIORef eqEpsilonRef x
 
+----------------------------------------------------------------------------
+-- Helper functions
+----------------------------------------------------------------------------
+
+go_assert_list :: (a -> a -> Assertion) -- ^ The function used to make an assertion on two elements (expected, actual)
```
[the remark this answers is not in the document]
sfindeisen: Fixed.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -1,25 +1,30 @@
...
+
 newtype EqEpsilon = EqEpsilon Rational
   deriving (Typeable, Num, Fractional)
 
 instance IsOption EqEpsilon where
   defaultValue = EqEpsilon eqEpsilonDefault
   parseValue s = fmap (EqEpsilon . toRational) ((safeRead :: String -> Maybe Double) s)
   optionName = return "eq-epsilon"
-  optionHelp = return $ "Epsilon to use for floating point comparisons: abs(a-b) < epsilon . Default: " ++ show eqEpsilonDefault
+  optionHelp = return $ "Epsilon to use for floating point comparisons: abs(a-b) < epsilon . Default: " ++ show (fromRational eqEpsilonDefault :: Double)
```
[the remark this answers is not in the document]
sfindeisen: Fixed :)

test/common/TestCommon.hs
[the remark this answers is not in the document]
Mikolaj: Correct, I guess. Yes, this concrete instance is useful and the shape would be legit. I suppose that's worth the risk for when we start using any of the other dozen of matrix types.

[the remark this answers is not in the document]
sfindeisen: When we introduce new matrix types (higher dimensions?...), we can (and should) add more instances. But even if we forget to do so, the worst thing that can happen is (I think) that we mistakenly judge some `2x3` and `1x6` matrices, miraculously containing the same data, to be equal. This is highly unlikely.

Do you have any specific matrix types in mind? Perhaps we could implement those instances now.

sfindeisen:
I only implemented the tuples up to `4`. More boilerplate, anyone? Up to `10`? `12`?

sfindeisen:
Speaking of those tuples.... perhaps we could implement `Linearizable` and `HasShape` instances for homogenous tuples (i.e. `(a,a,a,a)` etc.) which would be quite compact, and then not worry about heterogenous tuples longer than `4` until they're needed in practice?

test/common/TestCommon.hs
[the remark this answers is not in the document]
Mikolaj: Massiv and/or Accelerate matrices, as well as anything that may be needed for GPU. I think it's too early to touch the former two. I agree the risk is low. What bothered me was I didn't understand the need for that instance, but now I do. :)

Mikolaj:
> Speaking of those tuples.... perhaps we could implement `Linearizable` and `HasShape` instances for homogenous tuples (i.e. `(a,a,a,a)` etc.) which would be quite compact, and then not worry about heterogenous tuples longer than `4` until they're needed in practice?

The current examples, which I'd like to rewrite in the new style, have 7-tuples or more, IIRC. Not homogeneous. We may want to nest them somehow, but a 5-tuple sounds likely, anyway. I think up to 10, inclusive, would be good to have, just so that users feel provided for.

test/common/TestCommonEqEpsilon.hs
```diff
@@ -34,23 +40,245 @@ eqEpsilonRef = unsafePerformIO $ newIORef eqEpsilonDefault
...
+                   -> a      -- ^ The actual value
+                   -> Assertion
+assert_close_eps preface epilogue eqEpsilon expected actual = do
+  assertBool (message eqEpsilon) (abs (expected-actual) <= eqEpsilon)
+  where
+    msg = "expected: " ++ show expected ++ "\n but got: " ++ show actual
+    t = lowercase
+    message errorMargin = (if null preface then "" else preface ++ "\n") ++
+                          msg ++ "\n (maximum margin of error: " ++ show errorMargin ++ ")" ++
+                          (if (null epilogue) || (t epilogue == t preface) || (t epilogue == t msg) then "" else "\n" ++ epilogue)
```
Mikolaj, line 107: Thank you for printing the whole vector in full. That's very helpful.

PR #74, opened 2022-11-22
test: conv2d ad vs manual dInp and dKrn

tranma:
Hi @Mikolaj 
I've fixed the "manually differentiated" conv2d_dInp and dkrn and re-ran the tests against the ad version. Works 🎉 
Apologies for the confusion.

### DIFF
--- a/test/common/TestGradient.hs
+++ b/test/common/TestGradient.hs
@@ -9,6 +9,7 @@ import Prelude
 import qualified Data.Array.Convert
 import qualified Data.Array.ShapedS as OS
 import           GHC.TypeLits (KnownNat, type (+))
+import           Data.Proxy
 import           Numeric.LinearAlgebra (Numeric)
 import qualified Numeric.LinearAlgebra as LA
 import           Test.Tasty
@@ -20,6 +21,7 @@ import HordeAd.Internal.Delta (atIndexInTensor)
 
 import Tool.EqEpsilon
 import Tool.Shared
+import GHC.TypeLits
 
 testTrees :: [TestTree]
 testTrees = [ quickCheckForwardAndBackward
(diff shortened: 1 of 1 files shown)


Mikolaj:
Thank you very much! Shall I review the PR, asking questions inline, or would you prefer me to wait for the meeting (in which case, I'd just merge, then improve the code to the best of my knowledge and ask for forgiveness during the meeting:)?

tranma:
Yes we can just talk right here if you’d like. Feel free to review/comment inline

test/common/TestGradient.hs
```diff
@@ -288,28 +290,57 @@ testBarR =
 
 -- The following are borrowed from https://github.com/benl23x5/adops.
 
--- | Derivative of unpadded full convolution with respect to the input image.
+-- | Derivative of padded full convolution with respect to the input image.
 --
 conv2d_dInp
-  :: forall shK shA shB shK1 nImgs nCinpA nAh nAw nCoutK nCinpK nKh nKw r.
+  :: forall shK shA shB shB1 sh1 nAimg nAcha nAh nAw nKimg nKcha nKh nKw r.
```
Mikolaj, line 296: For our sanity, could we stick to the same dimension size names here and in DualNumber.hs? In DualNumber.hs, in turn, I tried to stick to the dimension names from

https://github.com/benl23x5/adops/blob/8c16600e58e1b4c8181daa6a28f0d9765c2f9d48/src/Adops/Op/Conv.hs#L6

as opposed to 

https://github.com/benl23x5/adops/blob/8c16600e58e1b4c8181daa6a28f0d9765c2f9d48/src/Adops/Op/Conv.hs#L124

but I'm fine with either choice, as long as it's consistent.

test/common/TestGradient.hs
```diff
@@ -337,22 +368,37 @@ dotOS :: (Numeric r, OS.Shape sh)
       => OS.Array sh r -> OS.Array sh r -> r
 dotOS arr1 arr2 = OS.toVector arr1 LA.<.> OS.toVector arr2
 
+sumOS :: (Numeric r, OS.Shape sh)
+      => OS.Array sh r -> r
+sumOS arr = sum $ OS.toList arr
+
```
Mikolaj, line 374: That's a fine definition, but it's already defined upstream as OS.sumA.

test/common/TestGradient.hs
```diff
@@ -288,28 +290,57 @@ testBarR =
...
+     , shA ~  '[nAimg, nAcha, nAh, nAw]
+     , shB ~  '[nAimg, nKimg, nAh, nAw]
+     , shB1 ~ '[1,     1,     nAh, nAw]
+     , sh1  ~ '[nKimg])
   => OS.Array shK r
   -> OS.Array shB r
   -> OS.Array shA r
 conv2d_dInp arrK arrB =
+  let nKh = fromIntegral (natVal $ Proxy @nKh) :: Int
+      nKw = fromIntegral (natVal $ Proxy @nKw) :: Int
```
Mikolaj, line 311: Well done. I guess the proxies are unavoidable in the padded version.

test/common/TestGradient.hs
```diff
@@ -337,22 +368,37 @@ dotOS :: (Numeric r, OS.Shape sh)
...
-        -- and fails to typecheck with the commented out dimension size,
-        -- because conv2d_dInp lacks the arrA parameter
-  in assertEqualUpToEpsilon 1e-7
-       (revDt (conv2d (constant arrK))  -- gradient wrt second argument
-              arrB   -- dummy, should be arrA, the point at which we take
-                     -- the gradient: https://en.wikipedia.org/wiki/Gradient
-              arrB )  -- sensitivity; should we normally test with 1?
-       (conv2d_dInp arrK arrB)  -- expected result
+  let -- Input of shape: batch x chas x height x width
+      arrA   = 1 :: OS.Array '[1, 2, 4, 4] Double
```
Mikolaj, line 383: A good pre-test, before I add a QuickCheck test based on your code, is to try setting all dimensions to be different, unless the names of dimension lengths are the same (or there is something like `nAcha ~ nKcha` in a function signature). We use several functions, so a couple more dimensions may have to be equated but this should be explainable. Here is where naming dimensions consistently across all functions helps a lot.

Mikolaj (commented):
Thank you for sharing this version promptly. The ability to ask questions and contribute to this PR is going to help me understand the code and the maths better than if you plopped a perfect over-tested blob that I can only accept in awed silence. :)

test/common/TestGradient.hs
```diff
@@ -288,28 +290,57 @@ testBarR =
 
 -- The following are borrowed from https://github.com/benl23x5/adops.
 
--- | Derivative of unpadded full convolution with respect to the input image.
+-- | Derivative of padded full convolution with respect to the input image.
```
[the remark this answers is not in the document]
tranma: In the adops repo, we had a comment saying this is an "unpadded convolution -- output shape is the same as the input shape." The comment is wrong because if convolution doesn't pad its input, then the output should be smaller. What it should really say is "No custom padding. Only pad the input with 0s such that output is the same shape as the input." I'll update both adops and this to be clearer.

test/common/TestGradient.hs
tranma: ok I'll make them consistent

test/common/TestGradient.hs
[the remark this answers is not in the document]
tranma: I've fixed the docs over in adops.

test/common/TestGradient.hs
Mikolaj: Thank you. That also makes it easier for me to spot what I don't understand. E.g., is it significant that in `conv2d_dInp` there is, in https://github.com/benl23x5/adops/blob/master/src/Adops/Op/Conv.hs

    Shape4 nImgs  nCoutB nBh nBw = shape arrB

but in `TestGradient.hs`

    shB  ~ '[nImgs, nCout, nAh, nAw]

(where the latter agrees with  `conv2d` in `DualNumber.hs` and in `Conv.hs`). Or am I badly pattern-matching this? Or is naming in horde-ad code made consistent only internally, not wrt adops code?

test/common/TestGradient.hs
```diff
@@ -344,15 +380,26 @@ adoptTests = testGroup "Tests of the port of adopt code"
...
-              arrB )  -- sensitivity; should we normally test with 1?
-       (conv2d_dInp arrK arrB)  -- expected result
+  let -- Input of shape: batch x chas x height x width
+      arrA   = 1 :: OS.Array '[5, 2, 4, 8] Double
+      -- Filters of shape: num_filters x chas x kernel_height x kernel_width
+      arrK   = 1 :: OS.Array '[7, 2, 1, 3] Double
+      -- Output gradient of shape: batch x chas x output_height x output_width
+      arrB'  = 1 :: OS.Array '[5, 7, 4, 8] Double
+      -- Compare the ad version against the manual derivative.
+      dInp   = conv2d_dInp arrK arrB'
```
Mikolaj, line 390: I still can't wrap my head around why conv2d_dInp doesn't take arrA as an argument. Does it assume it's 0, 1 or some other fixed value? At what point is the gradient computed?

test/common/TestGradient.hs
tranma: In Adops the shape is given in value-land. So we have a runtime check:
```
 = let  Shape4 nImgs  nCoutB nBh nBw = shape arrB
        Shape4 nCoutK nKchan nKh nKw = shape arrK
        nCout   = same nCoutB nCoutK -- blow up if no good
```
In horde-ad the shape is encoded in the type. To match Adops syntactically we could write:
```
forall nCoutB nCoutK....
shB  ~ '[nImgs, nCoutB, nAh, nAw],
shK  ~ '[nCoutK, nKchan, nKh, nKw],
nCoutB ~ nCoutK
```
or just use the same name `nCout` and get the same constraint result.

test/common/TestGradient.hs
tranma: Hi Mikolaj,
In the general case, ie for some arbitrary function f(x), we do need the point 'x' at which the gradient is defined. For example if I were to write a manual version of the gradient for f(x) = x^2:
```
sqrt x = x * x
sqrt_dInp x dOut = 2 * x * dOut -- the VJP of sqrt at x
```
But convolution is a linear function (in the sense of a linear combination) in all of its arguments: 'conv(input, filters)' depends on 'input' linearly and depends on 'filters' also linearly. We can see this in the definition for convolution at each index: it involves only a product of 'input' and 'filters' (then a sum).
A linear function has the same gradient at every point (eg f(x) = x ==> f'(x) = 1 for all x) so there is no need to take the arrA argument. This would not be true for so-called "non-linearities".
Hope this helps. We can chat more in the meeting!
Cheers
Tran

test/common/TestGradient.hs
Mikolaj: Yes, that makes sense; let's keep `nCout` as is. However, I was alluding to there being "nBh nBw" in the first case vs "nAh, nAw" in the second.

tranma: oh right, sorry I missed that. In this particular version of convolution we are padding the input such that the output's spatial dimensions are the same as the input's. Having it in the types is a nice bonus for horde!

test/common/TestGradient.hs
Mikolaj: Oh, great, thank you, I knew I would learn some useful maths by asking clueless questions just following syntax and types. That means I can ask QuickCheck to randomly generate arbitrary "point" arguments and test with them. If I knew this earlier, I could even test automatic derivative of convolution in isolation. I can't wait to apply QuickCheck to this code.

test/common/TestGradient.hs
Mikolaj: Oh, right, now I can see "the output size is the same as the input size", so that's in the comment, not in the variable names, while type-level requires it to be reflected everywhere. Puzzle solved.

BTW, could you kindly `git rebase master`? Then I'd merge immediately, before it gets out of sync again.

Mikolaj:
@tranma: Thank you very much for the PR. Merging after CI passes.

PR #77, opened 2022-12-02
Extend Tran tests of her conv2d gradients to QuickCheck

Mikolaj:
This extends to QuickCheck the tests, written by @tranma. They test Tran's manual gradients of conv2d vs gradients automatically computed by horde.

QuickCheck is testing this 100 times by default. Only small random arrays of up to 2.5K elements each are used, because horde-ad is not yet optimized for the build/index coding style (but for bulk operations instead).

This group of tests is run with 

    cabal test shortTestForCI --enable-optimization --test-options='-p "Tests of the port of adopt code" --quickcheck-replay=3

where the last number determines the main seed and can be changed.


Mikolaj:
Let me merge before the test harness bitrots completely. We can adjust, extend the test, fix the gradients, etc., later on.

PR #78, opened 2022-12-05
Add disparity cost volume example

benl23x5:


example/Disparity.hs
```diff
@@ -0,0 +1,93 @@
...
+ .  (Num a, R.UniformRange a, F.Storable a, OS.Shape sh)
+ => Int -> OS.Array sh a
+random seed
+ = let  xs = OS.fromList
+           $ take (OS.size xs)
+           $ unfoldr (Just . R.uniformR (-1, 1)) (R.mkStdGen seed)
+   in   xs
+
+
+-- TODO: where is the real version of this defined?
```
Mikolaj, line 90: Are you looking for this function?

https://github.com/Mikolaj/horde-ad/blob/master/test/common/TestGradient.hs#L359

It's defined here:

https://github.com/Mikolaj/horde-ad/blob/master/src/HordeAd/External/Adaptor.hs#L31-L36

example/Disparity.hs
```diff
@@ -0,0 +1,93 @@
...
+--   Described in:
+--    Anytime Stereo Image Depth Estimation on Mobile Devices
+--    Wang, Lai et al, ICRA 2019
+--    https://arxiv.org/abs/1810.11408
+--    Section III b).
+--
+costVolume
+ :: forall d r nImgs nChas nRows nCols nCount shA shO.
+    ( ADModeAndNum d r
+    , KnownNat nImgs, KnownNat nChas, KnownNat nRows, KnownNat nCols, KnownNat nCount
```
Mikolaj, line 39: For the one that has `StaticNat`, you don't need the `KnownNat` any more. The rule of thumb seems to be, if the compiler is giving you grief or if the function is expected to be used by other users, not just by your other function, then do `StaticNat` for all dimension arguments.

example/Disparity.hs
Mikolaj: And here's the low level bit:

https://github.com/Mikolaj/horde-ad/blob/master/src/HordeAd/Core/Engine.hs#L64

but you seem to have reimplemented it just fine?

The real thing from Adaptor let's you do `value f` and then apply it (multiple) arguments repeatedly, while the low level version requires you to first fully apply `f`.

example/Disparity.hs
```diff
@@ -0,0 +1,93 @@
...
+         [ "arrL  = " ++ show arrL
+         , "arrR  = " ++ show arrR
+         , "arrS  = " ++ show arrS
+         , "arrO  = " ++ show arrO
+         , "arrDR = " ++ show arrDR
+         , "arrDL = " ++ show arrDL ]
+
+
+-- | Generate an array of random values for testing.
+random
```
Mikolaj, line 79: This is a different implementation of

https://github.com/Mikolaj/horde-ad/blob/master/src/HordeAd/External/Adaptor.hs#L278

though yours may be more convenient by permitting type applications.

Mikolaj: I've just pushed a tiny change that probably makes `randomVals @'[1, 2, 4, 6] @Double` possible, just as in your reimplementation.

Mikolaj (approved):

example/Disparity.hs
```diff
@@ -0,0 +1,93 @@
...
+ :: forall sh a
+ .  (Num a, R.UniformRange a, F.Storable a, OS.Shape sh)
+ => Int -> OS.Array sh a
```
Mikolaj, line 82: BTW, this is pretty cool formatting. I may borrow it.

Mikolaj:
@benl23x5: any changes you'd like to make before merging?

We are experimenting with vectorizing `build` in the context of ranked tensors and such tasty nested `build` as you have here would be very useful to test and benchmark that.

Mikolaj:
I took the liberty of rebasing and I'm going to merge now. Thank you for the PR!

Mikolaj:
I've also recast the `costVolume` code in ranked tensors form and I added tests it:

https://github.com/Mikolaj/horde-ad/blob/51c8504443680dc6f86406e5c146207b80bb5b09/test/simplified/TestConvSimplified.hs#L538-L651

I can't find a closed form for the derivative of `costVolume` nor any handy properties of its result, so I can't add any QuickCheck tests.

PR #81, opened 2023-01-31
Include GHC.Nat-indexed lists by Tom Smeding

Mikolaj:


simplified/HordeAd/Core/Ast.hs
```diff
@@ -20,30 +23,174 @@ module HordeAd.Core.Ast
...
+deriving stock instance Functor (Index n)
+
+instance Show i => Show (Index n i) where
+  showsPrec _ Z = showString "Z"
+  showsPrec d (idx :. i) = showParen (d > 3) $
+    showsPrec 3 idx . showString " :. " . showsPrec 4 i
+
+-- I'm afraid, I can't do the unsafeCoerces below with this order
+-- of argument in :., can I? So I need this instead:
+pattern (:.) is i = S i is
```
Mikolaj, line 64: @tomsmeding, at your leisure, if you could have  alook at 2 problems I've encountered so far and my partial solutions. This is the first.

simplified/HordeAd/Core/Ast.hs
```diff
@@ -20,30 +23,174 @@ module HordeAd.Core.Ast
...
+-- enough to make mistakes even with this dumb wrapper, so it might be fine.
+newtype Shape n i = Shape (Index n i)
+  deriving Show
+
+-- It seems that function names can't start with colon. That's too bad.
+-- Also, I can't make pattern synonym of out that because newtype is in the way.
+-- Or can I if I defined two pattern synonyms?
+infixl 3 @$
+(@$) :: Shape n i -> i -> Shape (1 + n) i
+Shape sh @$ s = Shape (sh :. s)
```
Mikolaj, line 96: @tomsmeding: this is the second.

simplified/HordeAd/Core/Ast.hs
```diff
@@ -31,19 +38,138 @@ import           GHC.TypeLits (KnownNat, Nat, type (+))
...
+
+dropIndex :: forall len n i. KnownNat len
+          => Index (len + n) i -> Index n i
+dropIndex ix = unsafeCoerce $ drop (valueOf @len) $ unsafeCoerce ix
+
+permutePrefixIndex :: [Int] -> Index n Int -> Index n Int
+permutePrefixIndex p ix =
+  let l = unsafeCoerce ix
+  in (unsafeCoerce :: [Int] -> Index n Int)
+     $ V.toList $ VS.fromList l V.// zip p l
```
tomsmeding, line 90: Please check that `p` has the right length, otherwise this is wildly unsafe (the `zip` will stop at the end of the shortest list)

simplified/HordeAd/Core/Ast.hs
tomsmeding: View patterns and explicitly bidirectional pattern synonyms are way too powerful.
```haskell
pattern (::.) :: Shape n i -> i -> Shape (1 + n) i
pattern sh ::. i <- ((\(Shape (sh :. i)) -> (Shape sh, i)) -> (sh, i))
  where pattern (Shape sh) ::. i = Shape (sh :. i)
```
I hope I did that correctly, don't have a compiler handy.

simplified/HordeAd/Core/Ast.hs
Mikolaj: Well spotted.

simplified/HordeAd/Core/Ast.hs
tomsmeding: I get what you are doing with the unsafeCoerces, and that's horrible. :D In particular I'd ask you to add a strongly worded comment above the definition of `Index` imploring the reader to not change the representation of the data type; for example, adding an `n >= 0` constraint to the `(:.)` case is already wrong if I'm not mistaken, because that becomes an additional field in the data type, breaking representational equality with lists. Or so I've been told — happy to be corrected here.

And no, if you want representational equality with lists, then you cannot change the, well, representation, so no way around using a pattern synonym to fix the argument order. :)

Mikolaj: Agreed. I will try to add a strongly worded comment that is still family-friendly.

tomsmeding: I wonder if you really need the unsafeCoerces, though. `take` on a list js not going to be faster than reimplementing the same directly on `Index`; indeed, as we've determined, the two have the same representation so hopefully we'll even get the exact same code out in the end. Same for `tail` and `drop`.

For permutations, `VS.fromList` has to walk the list anyway, so reimplementing that for `Index` should again result in the same code. Then, `VS.toList` could be reimplemented by just indexing into the output vector — that should be the same as what fromList is doing now.

And here we have eliminated all unsafeCoerce, I believe, meaning that you can change the argument order of the second constructor of `Index` back to what it should be! :)

Mikolaj: Haha, you've elided the tiny detail of actually writing `take` without `unsafeCoerce` using non-unary Nat. And the same for any future list function we realise we need. I did try to write `take` in a honest way first...

tomsmeding: Do you happen to know why [this thing](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/GHC.TypeNats.html#withSNat) is not exported? It would be nice here.

Also did I mention that I hate GHC Nats?

But you're right of course. Still I wonder whether the following would incur any runtime penalty:
```haskell
takeIndex :: forall len n i. KnownNat len
          => Index (len + n) i -> Index n i
takeIndex idx =
  let len = natVal (Proxy @len)
  in if len < 0 then error "wat" else go len idx
  where
    go :: Integer -> Index m i -> Index m' i
    go 0 idx' = unsafeCoerce idx'
    go len Z = error "wat"
    go len (S _ idx') = go (len - 1) idx'
```
If not, that would allow at least avoiding the `(:.)` pattern synonym (because we can fix the argument order in the data type).

Mikolaj: > Do you happen to know why [this thing](https://hackage.haskell.org/package/base-4.16.0.0/docs/src/GHC.TypeNats.html#withSNat) is not exported? It would be nice here.

No, I don't know. Probably authors thought this bit is less stable than the rest. But if so, it should be in an .EvenMore.Internal module. Perhaps they can be persuaded `withSNat` itself seem pretty solid and essential, unlike the hacks used to implement it. Analogous things are certainly provided and exposed for other Nat implementations.

Re `nonsense`, where would the runtime penalty come from? I don't see anything suspect.

Re `(:.)` pattern synonym, perhaps let's keep it? I have a dream that one day we'd have such a sized list in base and then I'd like to just effortlessly replace our implementation with it, keeping the synonym and a few specialized functions (and `Shape`).

simplified/HordeAd/Core/Ast.hs
Mikolaj: Fixed.

simplified/HordeAd/Core/Ast.hs
tomsmeding: Hm, fair point, perhaps having the pattern synonym doesn't really bring any downsides. (I don't have particularly high hopes for a suitable sized list type in base, but who knows.)

I also don't think `nonsense`-style functions would bring a runtime penalty, but you have more knowledge of such matters than I do, hence I asked :)
The advantage of this style is being independent of the fragile equality of representation between `Index n` and `[]`; it doesn't save any `unsafeCoerce`. The disadvantage is clearly verbosity.

Mikolaj:
@tomsmeding: while traceShow-debugging, I've stumbled on something shocking:

> The shape of an array, i.e., a list of the sizes of its dimensions. In the linearization of the array the outermost (i.e. first list element) varies most slowly. O(1) time.
https://hackage.haskell.org/package/orthotope-0.1.2.0/docs/Data-Array-ShapedS.html#v:shapeL

So the order of indexes I'm using, aping orthotope, seems to be inverse what we have on the sized list. Consequently my strides-based index computation gives different results than `fromLinearIdx` applied to corresponding shapes. Probably other things would differ, too, but I only tweaked and tested this single part.

The shocking part is that only one test failed, only sometimes, and never Doubles, only with Floats. Apparently it all somehow symmetrically cancelled out.

What do we do?

tomsmeding:
> only one test failed

That's interesting, seems to be some kind of symmetry in your tests

> ... only sometimes, and never with Doubles, only with Floats

... and seemingly said symmetry is only there mathematically, not numerically. Perhaps? This is kind of wild though.

In any case, that haddocks description seems to match with how we represent things in a `Shape n i` (with the `:.` order, not the `S` order): 1. they are _sizes_, not strides, and 2. the left-most varies most slowly (and hence the right-most the most quickly). It seems the author of `orthotope` and I were in agreement all along. If I understand correctly, you were assuming something different in the rest of the code; is that right?

Mikolaj:
You write "the index 'Z :. i :. j' represents 'a[i][j]' in -- traditional C notation." And in C notation `i` is the outermost dimension."

```
int test[2][3][4] = {
    {{3, 4, 2, 3}, {0, -3, 9, 11}, {23, 12, 23, 2}},
    {{13, 4, 56, 3}, {5, 9, 3, 5}, {3, 1, 4, 9}}};```

**Comment by Mikolaj** [comment]
I think we have the same order as orthotope and C visually, but the reverse order vs orthope list-nesting-wise.

**Comment by Mikolaj** [comment]
But perhaps ours is better or more standard?

**Comment by tomsmeding** [comment]
> I think we have the same order as orthotope and C visually, but the reverse order vs orthope list-nesting-wise.

Precisely that! I'm not sure which list-nesting is better. Do you usually batch your operations by vectorising them, i.e. letting them work on large tensors instead of scalars? (If so, use the orthotope slowest-at-the-outside nesting.) Or do you usually batch your operations by mapping the operation over a large tensor of jobs? (If so, use my fastest-at-the-outside nesting, i.e. current `Shape n i`.)

Accelerate uses the fastest-at-the-outside nesting because the whole language is in this "intermediate" (in The Doc parlance) form where you can't have vectors of vectors, so all operations are rank-polymorphic instead; you get types like these:

```haskell
fold :: (Shape sh, Elt a) => (Exp a -> Exp a -> Exp a) -> Exp a -> Acc (Array (sh :. Int) a) -> Acc (Array sh a) 
```

this is a fold over the _innermost_ (fastest-moving) dimension. The "rank-polymorphic" bit is the polymorphism in `sh`. (Note that Accelerate's `Shape` is defined as `data sh :. i = sh :. i ; data Z = Z`, where usually `i ~ Int`.) But this is an Accelerate characteristic, no need to copy this if your operations work the other way round.

Mikolaj:
That's fascinating. For better or worse, I do the same as orthotope, that is, manipulate the outermost dimension and almost never fold or map over scalars. What Andrew proposes is also not focusing on scalars, but building or mapping over arbitrary tensors, which are nested in arbitrary larger tensors. And in case of `Sum`, keeping the current semantics that sums over the outermost dimension only and in `Sum0`, getting down from top to the very scalar. No variant that would sum the innermost dimension, leaving others alone. I have no perspective to tell to what extent this started via the influence of orthotope or more generally the highly parallel hardware both Lennart and Andrew were/are working on and to what extent this has anything to do with POPL AD work. But it seems to work well with the vectorization route we're following.

In the old code there is exactly one place where I reverse the index list: that's where I vectorize not by pushing things from the outermost dimension down, but by analysing special cases, such as build (\ix -> index v (ixs :. i)) where if i == ix then the result is slice (index v ixs), etc.

tomsmeding:
Then it seems that for your purposes, you should reverse `Index` (and `Shape`) :) I guess that also relieves you of the necessity of having the `(:.)` pattern synonym, because now the list is in natural cons-order.

Mikolaj:
Actually, I've now confused myself completely. The orthotope shapes that work best for me have the outermost (slowest) dimension least nested. But the index lists I'm using in the old code and in `index` and in `build` seem to have exactly the reverse nesting. Oh dear. I hope I panicked and it's actually consistent.

tomsmeding:
You see? You should have used `Index n` from the beginning!

Mikolaj:
I knew I shouldn't risk typing my code.

Mikolaj:
I've just checked and orthotope is consistent: in the few places that they use index lists, they do this in the same nesting as shapes. The uses are in `build` that takes a function from index lists to scalars and `update`, that updates a scalar at a given index list. I'm also consistent in my oldest code (not `simplified`) where I just follow orthotope. I may also be consistent in parts of `simplified` (build operation and a few others), but not in vectorization (and so in `index`, `indexN`), where I reverse the order, because it's so much more convenient in the scenario where you push indexing down the tensor expression and always have tower of indexes above you and a multi-dimensional tensor below you. With the inconsistent order, I have the closest dimesions of the tensor below in in the head of shape, and the closest index above me in the head of the index list.

So it's only me. And I have to decide on just one scheme. For shapes definitely orthotope scheme works great both for vectorized operations and for the vectorization process. I guess the order of indexes has to change in my newest code to match that

So, I agree, let's reverse our sized list and get rid of the pattern. I wonder if that's going to change any of the implementations. In any case, we will also need a `snoc`, `last` and `init` operations for indexs, just as the old `cons`, `head` and `tail` for shapes. What was the about finger trees? ;(

tomsmeding:
Finger trees are what `Data.Sequence` is using, i.e. something that you can cons and snoc on in amortised O(1) time :)

It's also far too complicated to even think of making a Nat-indexed version of it --- not because it's actually super complicated ([this functional pearl](https://www.cs.tufts.edu/~nr/cs257/archive/koen-claessen/finger-trees.pdf) gives a very nice explanation), but because making Nat-indexed versions of things gets _very_ hairy _very_ quickly.

> So it's only me. And I have to decide on just one scheme. For shapes definitely orthotope scheme works great both for vectorized operations and for the vectorization process. I guess the order of indexes has to change in my newest code to match that

That, or have two `Index` types, one for each direction. Depending on how much the reversed order helps your code there, it might be worth the additional administration of having two types.

Mikolaj:
Yeah, I'm pretty sure I want to switch to orthotope nesting in all places. So just the reverse of our sized lists right now. I'm just rewriting `fromLinearIdx` randomly until the Float test succeeds (no luck so far).

tomsmeding:
Untested, so not sure if I'm better than your "no luck so far", but this is what I would write: https://paste.tomsmeding.com/f3UVMR5f

tomsmeding:
Good fix! Of course one should return `n` in the `go [] n` case. I was halfway writing a response indicating why your fix was wrong when I finally figured out that you are correct.

I'm still interested by the test symmetries. Surely, if those symmetries end up working so well, then you must be very consistently (in some perverted way) doing the orderings wrong, or sometimes right! Anyway, I hope we don't even need to really understand the reason behind this. :)

simplified/HordeAd/Core/Ast.hs
Mikolaj: With the simplified (:.) that is no longer a pattern synonym, I've almost made it to work and it's so useful and improves safety a lot. I think the only wrong detail is the type of the new pattern, which then causes these to fail:

```
  AstTranspose v -> case shapeAst v of
    Shape (i :. k :. sh) -> k :$ i :$ Shape sh
    sh -> sh  -- the operation is identity if rank too small

-- even this fails (probably the type of (:$) is not hairy enough):
--    ZS -> ZS
--    a :$ bs -> a :$ bs

-- this is how it should look:
--    i :$ k :$ sh -> k :$ i :$ sh
--    sh -> sh  -- the operation is identity if rank too small
```

tomsmeding: This seems to work? Up to you to fix code style and organisation https://paste.tomsmeding.com/IKwLCzCk

Mikolaj: Mind blown. Thank you.

tomsmeding: The thing is that your type was correct and mine was not (indeed `n` should be existential, and the equality constraint should be provided -- this is precisely what you wrote). However, the code didn't typecheck under this correct type signature because the type of that inline lambda would be very weird: it would return something of existential type, which is dodgy.

So I "simply" added an explicit type signature to that lambda (and extracted it to a top-level function for sanity reasons), encoding the existential in a data type; and then we still need to export the proof that relates the existential `n` to the original `n1`, so we export that too. Then you get this code.

simplified/HordeAd/Core/Ast.hs
tomsmeding: (My knowledge here comes from writing an early incorrect version of, and later understanding the actually correct version of, [this code here](https://github.com/AccelerateHS/accelerate/blob/5971c5d8e4dbba28d2017e7ce422cf46a20197cb/src/Data/Array/Accelerate/AST/Idx.hs#L65-L92))

simplified/HordeAd/Core/Ast.hs
tomsmeding: (My knowledge comes from writing an early incorrect version of, and later understanding the actually correct version of, [this code here](https://github.com/AccelerateHS/accelerate/blob/5971c5d8e4dbba28d2017e7ce422cf46a20197cb/src/Data/Array/Accelerate/AST/Idx.hs#L65-L92))

Mikolaj: "Simply" is well put.

simplified/HordeAd/Core/Ast.hs
tomsmeding: Turns out other people thought the same about `SNat` and friends, so [in GHC 9.6](https://github.com/haskell/core-libraries-committee/issues/85) this will change :)

Mikolaj: Oh, cool. Do I understand correctly that this would enable a purer implementation of some or our functions, but it does not break our dirty code, at least not yet in GHC 9.6?

Mikolaj:
@tomsmeding: unless you find any hidden traps, the PR is done. The 3 tests that did "arithmetic underflow" now work fine (which may or may not be the result of fixing a one-liner one-off dimension bug revealed by sized indexes). Next steps will be generalizing build, gather and scatter to function from sized lists of variables to Index (Int to Index currently) and simplifying some Tensor class method signatures now that we can express them sanely. Then we can complete vectorization (in particular the case of vectorizing gather). But all that's outside the scope of this PR.

I'm still not quite sure I was consistently using the reversed order of indexes vs shapes previously. But that's not impossible and that would explain a part of the mystery of the same tests passing with both. In any case, now I'm pretty sure we are consistently using the other order of indexes throughout the code. And we eliminated 90% of unsafeCoerce in vectorization functions (and the 10% will go soon).

tomsmeding (approved):
The `SizedList` and `SizedIndex` modules I checked thoroughly, the rest I just skimmed or skipped entirely.

CREDITS.md
```diff
@@ -14,7 +14,7 @@ Edward Kmett, author of the archetype Haskell https://github.com/ekmett/ad libra
 
 Justin Le, author of the https://github.com/mstksg/backprop library and the amazing articles at https://blog.jle.im
 
-Tom Smeding, who influenced the major implementation decisions of horde-ad
+Tom Smeding, who influenced the major implementation decisions of horde-ad and solved the hardest technical problems
```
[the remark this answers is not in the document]
Mikolaj: True story.

src/common/HordeAd/Internal/SizedList.hs
```diff
@@ -0,0 +1,158 @@
...
+
+instance Show i => Show (SizedList n i) where
+  showsPrec _ Z = showString "Z"
+  showsPrec d (i ::: ix) = showParen (d > 3) $
+    showsPrec 4 i . showString " ::: " . showsPrec 3 ix
+
+deriving stock instance Functor (SizedList n)
+
+instance KnownNat n => Foldable (SizedList n) where
+  foldr f z l = foldr f z (sizedListToList l)
```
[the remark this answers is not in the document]
Mikolaj: Well spotted. Fixed.

src/common/HordeAd/Internal/SizedList.hs
```diff
@@ -0,0 +1,158 @@
...
+dropSized :: forall len n i. KnownNat len
+          => SizedList (len + n) i -> SizedList n i
+dropSized ix = unsafeCoerce $ drop (valueOf @len) $ unsafeCoerce ix
+
+unsnocSized :: SizedList (1 + n) i -> (SizedList n i, i)
+unsnocSized Z = error "unsnocSized: impossible pattern needlessly required"
+unsnocSized (i ::: ix) = case ix of
+  Z -> (Z, i)
+  _ ::: _ -> let (init1, last1) = unsnocSized ix
+            in (i ::: init1, last1)
```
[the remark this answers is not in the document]
Mikolaj: fixed

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+      in i :. sh
+  | LTI <- cmpNat (Proxy @1) (Proxy @n) =
+      let sh = listToIndexProxy @(n - 1) is
+      in i :. sh
+  | otherwise =
+      error "listToIndex: list too long"
+-}
+
+listToIndex :: forall n i. KnownNat n => [i] -> Index n i
+listToIndex = Index . listToSized
```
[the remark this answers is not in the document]
Mikolaj: I probably was applying types and wanted to make sure the order of type variables is fixed. But it works fine without and I'm no longer changing the type signature. Fixed.

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+  forall n. UnconsIndexRes (Index n i) i (Dict (n1 ~ (1 + n)))
+unconsIndex :: Index n1 i -> Maybe (UnconsIndexRes i n1)
+unconsIndex (Index sh) = case sh of
+  i ::: sh' -> Just (UnconsIndexRes (Index sh') i Dict)
+  Z -> Nothing
+
+deriving stock instance Functor (Index n)
+
+instance KnownNat n => Foldable (Index n) where
+  foldr f z l = foldr f z (indexToList l)
```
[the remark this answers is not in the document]
Mikolaj: Done.

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+  forall n. UnconsShapeRes (Shape n i) i (Dict (n1 ~ (1 + n)))
+unconsShape :: Shape n1 i -> Maybe (UnconsShapeRes i n1)
+unconsShape (Shape sh) = case sh of
+  i ::: sh' -> Just (UnconsShapeRes (Shape sh') i Dict)
+  Z -> Nothing
+
+deriving stock instance Functor (Shape n)
+
+instance KnownNat n => Foldable (Shape n) where
+  foldr f z l = foldr f z (shapeToList l)
```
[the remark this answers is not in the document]
Mikolaj: Removed.

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+
+singletonShape :: i -> Shape 1 i
+singletonShape = Shape . singletonSized
+
+tailShape :: Shape (1 + n) i -> Shape n i
+tailShape (Shape ix) = Shape $ tailSized ix
+
+takeShape :: forall m n i. KnownNat m
+          => Shape (m + n) i -> Shape m i
+takeShape (Shape ix) = Shape $ takeSized ix
```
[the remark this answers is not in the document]
Mikolaj: In my code it's always `m`, so it's easier to compare. But in general, it may be anything, so `len` is neutral and descriptive in the module we aspire to include in `base`.

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+shapeSize :: Num i => Shape n i -> i
+shapeSize ZS = 1
+shapeSize (n :$ sh) = n * shapeSize sh
+
+flattenShape :: Num i => Shape n i -> Shape 1 i
+flattenShape = singletonShape . shapeSize
+
+-- Warning: do not pass a list of strides to this function.
+listShapeToShape :: forall n i. KnownNat n => [i] -> Shape n i
+listShapeToShape = Shape . listToSized
```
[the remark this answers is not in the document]
Mikolaj: Removed.

src/common/HordeAd/Internal/TensorOps.hs
```diff
@@ -29,8 +31,8 @@ import           Numeric.LinearAlgebra (Matrix, Numeric, Vector)
 import qualified Numeric.LinearAlgebra as LA
 import qualified Numeric.LinearAlgebra.Devel
 import           Text.Show.Functions ()
-import           Unsafe.Coerce (unsafeCoerce)
```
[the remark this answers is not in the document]
Mikolaj: \o/

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+
+infixr 3 :.
+pattern (:.) :: forall n1 i. forall n. (1 + n) ~ n1
+             => i -> Index n i -> Index n1 i
+pattern i :. sh <- (unconsIndex -> Just (UnconsIndexRes sh i Dict))
+  where i :. (Index sh) = Index (i ::: sh)
+{-# COMPLETE ZI, (:.) #-}
+
+data Dict c where
+  Dict :: c => Dict c
```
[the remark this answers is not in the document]
Mikolaj: Tests pass fine. Thank you.

src/common/HordeAd/Core/SizedIndex.hs
```diff
@@ -0,0 +1,243 @@
...
+toLinearIdx :: forall m n i. Num i => Shape (m + n) i -> Index m i -> i
+toLinearIdx = \sh idx -> snd (go sh idx)
+  where
+    -- Returns (shape size, linear index)
+    go :: forall m1 n1. Shape (m1 + n1) i -> Index m1 i -> (i, i)
+    go sh ZI = (shapeSize sh, 0)
+    go (n :$ sh) (i :. idx) =
+      let (restsize, lin) = go sh idx
+      in (n * restsize, i * restsize + lin)
+    go _ _ = error "toLinearIdx: impossible pattern needlessly required"
```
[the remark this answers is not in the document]
Mikolaj: Perfect. Tests pass. Swapped.

src/common/HordeAd/Internal/TensorOps.hs
```diff
@@ -128,31 +136,29 @@ tindexR
...
+           Data.Array.Internal.T{offset, values}) = OR.normalize arr
       !_A = assert (offset == 0) ()
-  in let pathToIx is = sum (zipWith (*) is strides)
-         ix = pathToIx ixs
-         shN = drop (length ixs) sh
+  in let i = toLinearIdx @m @n (listShapeToShape sh) ix
+         shN = drop (valueOf @m) sh
          len = product shN
-     in OR.fromVector shN $ V.slice ix len values
+     in OR.fromVector shN $ V.slice i len values
```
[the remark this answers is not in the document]
Mikolaj: You are probably right on all accounts. Let's test...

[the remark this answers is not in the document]
Mikolaj: No, it seems you were mistaken about it being wrong. Tests pass. Swapped. I wonder if 

https://github.com/Mikolaj/horde-ad/blob/8e49146b893ad1ed238722326924c9688079ce5f/src/common/HordeAd/Internal/TensorOps.hs#L88-L102

can be simplified and sped up similarly.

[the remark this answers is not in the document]
Mikolaj: BTW, the new implementation of `tindexNR` if funny in that when projecting a 1M-element tensor to a single Double (presented as a rank 0 tensor), we keep the whole underlying 1M `values` vector. But this is fine --- the vector is going to be shared between all such Doubles and, probably, may even be a perfect packed representation for a set of all of them, as opposed to allocating each, separately, on the heap (even as a Double, not a rank 0 tensor).

That differs from `updateNR, where we actually update the `values` vector, which rewrites is completely, so not even a portion of the original vector is shared.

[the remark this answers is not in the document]
tomsmeding: It works?! Yay for strided representations, I guess.

Regarding updateNR: will those vjoin operations be done in-place somehow? If not, I think re-allocating the entire tensor after every update will be a much bigger inefficiency than the flattining, which is just a _single_ re-allocation, after all.

I wonder if you could allocate an `m`-dimensional array of `Int`s, where `-1` indicates "keep as-is" and anything `>=0` indicates the index in the `upd` list of the tensor to put there. Then `build` the destination array in one go. The boolean array seems redundant allocation but it's much less than the current _repeated_ reallocation of the entire thing.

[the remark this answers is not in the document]
tomsmeding: With the new non-allocating `tindexNR` it might be useful to export something like [ByteString.copy](https://hackage.haskell.org/package/bytestring-0.11.4.0/docs/Data-ByteString.html#v:copy) (semantically `id`) to destroy the sharing you mention. Sometimes it may be counterproductive.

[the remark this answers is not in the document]
Mikolaj: > Regarding updateNR: will those vjoin operations be done in-place somehow?

Right, they should be and probably will be in the future. But for now, I'm forgetting this is a list of updates and thinking about the simplest case of one-element list (which is the only one I'm using in horde-ad so far IIRC, incidentally). But you are right, context matters. Perhaps I should define a separate operation with just one update and it would be very similar to `tindexNR` and then try to optimize that separately.

Yes, using build for that makes some sense, especially for larger lists of larger tensors.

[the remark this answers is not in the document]
tomsmeding: Alternative to what I wrote above: or sort the `upd` list and construct one big list to be passed to `vjoin` _once_. That's actually probably the fastest one if the input arrays are sufficiently large and `upd` is not too long -- a bit of list manipulation is negligible compared to conditional code run elementwise over a huge array.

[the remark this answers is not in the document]
Mikolaj: Thanks. This may work. It's so fortunate subtensors can't overlap non-trivially. I will remember to look into these notes once I profile and see which  `updateNR` actually matters: the singleton one, or the general one.

PR #110, opened 2024-10-09
Use less Internal imports from ox-arrays

tomsmeding:
I exported some additional things from ox-arrays, and you were using some things from internal modules that were already exported.

https://git.tomsmeding.com/ox-arrays/diff/?h=singletons&id=2a3e58084da47d1dd3b557997e239ded480539a6&id2=48bb58de893bdd28a8bae534bf618fb5790833ca

There are less internal imports from ox-arrays now, but there are still a bunch left. When the decisions of whether to promote something to a public export or not started to become more difficult, I just copped out and stopped the process.

Also CI seems broken?


Mikolaj:
Thank you so much. Yes, the codebase is in turmoil, so it's possible it will stop using many of the left-over internal imports, so let's be lazy. :)

Re CI, yes, it's broken for a long time, since there is no GHC version that can compile it (but HEAD should be able to soon, I hope, and probably GHC 9.10.2 will, too).

PR #112, opened 2025-03-25
Updates to README.md

awf:
Some comments/proposed changes to new README.  May not adhere to Haskell naming conventions, or to be downright incorrect, but hopefully useful as nots.


Mikolaj (approved):
Andrew, thank you very much, your perspective is invaluable. I will merge this soon and then amend as needed. Before that, I've added two clarifying questions.

@tomsmeding: would you like to have quick look, too?

README.md
```diff
@@ -113,8 +139,8 @@ convMnistTwoS
   ( weightsReadout   -- readout layer weights, shape [10, n_hidden]
   , biasesReadout )  -- readout layer biases [10]
   =                  -- -> output classification, shape [10, batch_size]
-  gcastWith (unsafeCoerceRefl :: Div (Div w 2) 2 :~: Div w 4) $
-  gcastWith (unsafeCoerceRefl :: Div (Div h 2) 2 :~: Div h 4) $
+  gcastWith (unsafeCoerceRefl :: Div (Div w 2) 2 :~: Div w 4) $ -- ensure w divisible by 4 - what is :~:?
+  gcastWith (unsafeCoerceRefl :: Div (Div h 2) 2 :~: Div h 4) $ -- similar ?
```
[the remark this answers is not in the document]
awf: Yes, I think 
```hs
assumEquality (Div (Div w 2) 2) (Div w 4) $
```
would be nice!

README.md
```diff
@@ -94,11 +112,19 @@ A shorthand that creates the symbolic derivative program, simplifies it and inte
...
 (sfromListLinear [2,2] [2.4396285219055063,2.4396285219055063,2.4396285219055063,2.4396285219055063],sfromListLinear [2,2] [-1.953374825727421,-1.953374825727421,-1.953374825727421,-1.953374825727421],sfromListLinear [2,2] [0.9654825811012627,0.9654825811012627,0.9654825811012627,0.9654825811012627])
 ```
-
+AWF note: can we separate (creation and simplification) from interpretation, i.e. do that work once before applying many times?
+```hs
+>>> foograd = grad (kfromS . ssum0 . fooLet)  -- do all simplification etc here
+>>> foograd threeSimpleMatrices -- just interpret here?
+>>> foograd threeOtherSimpleMatrices
+```
+or maybe that happens automatically?
```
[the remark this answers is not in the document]
awf: I certainly think it's worth calling out that this is possible for fixed shapes.

awf:
> I will merge this soon and then amend as needed.
Or feel free to just co-edit this PR rather than merging the changes directly...

README.md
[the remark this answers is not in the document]
awf: Ah, I think I get it now, would it be something like this?
```hs
assumeEquality @(Div (Div w 2) 2) @(Div w 4) $
```
If so, yes, I think that's nice and clear.

README.md
```diff
@@ -34,25 +34,41 @@ which can be verified by computing the gradient at `(1.1, 2.2, 3.3)`:
...
-When `foo` is instantiated to matrices, which is a similarly trivial example as before due to the arithmetic operations working on the arrays element-wise, the gradient is:
+We can instantiate `foo` to matrices; the operations within (`sin`, `+`, `*`) etc applying elementwise:
 ```hs
-type Matrix2x2 f r = f (TKS '[2, 2] r)
+type Matrix2x2 f r = f (TKS '[2, 2] r) -- AWF: TKS means [...?]
 type ThreeMatrices r = (Matrix2x2 Concrete r, Matrix2x2 Concrete r, Matrix2x2 Concrete r)
 threeSimpleMatrices :: ThreeMatrices Double
-threeSimpleMatrices = (srepl 1.1, srepl 2.2, srepl 3.3)
-gradFooMatrix :: (Differentiable r, GoodScalar r)
+threeSimpleMatrices = (srepl 1.1, srepl 2.2, srepl 3.3) -- srepl duplicates its argument in each matrix entry
```
tomsmeding, line 42:
```suggestion
threeSimpleMatrices = (srepl 1.1, srepl 2.2, srepl 3.3) -- srepl replicates its argument to fill the whole matrix
```

README.md
```diff
@@ -34,25 +34,41 @@ which can be verified by computing the gradient at `(1.1, 2.2, 3.3)`:
...
+foo_value :: Matrix2x2 Concrete Double = foo threeSimpleMatrices
```
tomsmeding, line 43: If we define this, we should also show its value

README.md
```diff
@@ -34,25 +34,41 @@ which can be verified by computing the gradient at `(1.1, 2.2, 3.3)`:
...
+```
+We can compute a gradient of a function from matrices to scalar, say `sumfoo`:
```
tomsmeding, line 45:
```suggestion
Instantiated to matrices, `foo` now returns a matrix, not a scalar &mdash; but a gradient can only be computed of a function that returns a scalar.
Thus, let's define `sumfoo` which sums the output of `foo`: (note that `ssum0` returns a zero-dimensional array; `kfromS` extract the (single) scalar from that)
```

README.md
```diff
@@ -34,25 +34,41 @@ which can be verified by computing the gradient at `(1.1, 2.2, 3.3)`:
...
+```hs
+sumfoo :: ThreeMatrices r -> r
+sumfoo = kfromS . ssum0 . foo
+>>> sumfoo threeSimpleMatrices
+1.23456789 ????
+```
+and its gradient is given by
```
tomsmeding, line 52:
```suggestion
This function we can compute the gradient of as follows:
```
(also fix the indentation of the `=>` line in `gradsumfoo`)

README.md
```diff
@@ -34,25 +34,41 @@ which can be verified by computing the gradient at `(1.1, 2.2, 3.3)`:
...
 ```
 
-Note that `w` is processed only once during gradient computation and this property of sharing preservation is guaranteed for the `cgrad` tool universally, without any action required from the user. When computing symbolic derivative programs, however, the user has to explicitly mark values for sharing using `tlet` with a more specific type of the objective function, as shown below.
+[AWF: better explain `GoodScalar` name somewhere here]
+
+### Efficiency: preserving sharing
 
+We noted above that `w` appears twice in `foo`.  A property of tracing-based AD systems is that such re-use may not be captured, with explosive results.
+In `cgrad`, such sharing is preserved, so `w` is processed only once during gradient computation and this property is guaranteed for the `cgrad` tool universally, without any action required from the user.
+When computing symbolic derivative programs, however, the user has to explicitly mark values for sharing using `tlet` with a more specific type of the objective function, as shown below.
```
tomsmeding, line 71:
```suggestion
`horde-ad` also allows computing _symbolic_ derivative programs: using this API, a program is differentiated only once, after which it can be run on many different input values.
In this case, however, sharing is _not_ automatically preserved, so shared variables have to be explicitly marked using `tlet`, as shown below in `fooLet`.
This also makes the type of the function more specific: it now does not work on an arbitrary `Num` any more, but instead on an arbitrary `horde-ad` tensor that implements arithmetic operations like `atan2H`.
```
How should we explain the difference between `atan2` and `atan2H`?

README.md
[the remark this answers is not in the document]
awf: Just checking: if `threeOtherSimpleMatrices` are indeed the same shape, will the simplification be done only once?  If so, that's a good feature, which systems like JAX and PyTorch need to implement with explicit caching.  On the other hand, their caches can persist between invocations.

README.md
```diff
@@ -94,11 +112,19 @@ A shorthand that creates the symbolic derivative program, simplifies it and inte
...
+>>> foograd threeSimpleMatrices -- just interpret here?
+>>> foograd threeOtherSimpleMatrices
+```
+or maybe that happens automatically?
 
 ## For all shapes and sizes
 
-An additional feature of this library is a type system for tensor shape arithmetic. The following code is a part of convolutional neural network definition, for which horde-ad computes the gradient of a shape determined by the shape of input data and of initial parameters. The compiler is able to infer a lot of tensor shapes, deriving them both from dynamic dimension arguments (the first two lines of parameters to the function) and from static type-level hints. Look at this beauty.
+An additional feature of this library is a type system for tensor shape arithmetic. The following code is a part of convolutional neural network definition, for which horde-ad computes the gradient of a shape determined by the shape of input data and of initial parameters. The compiler is able to infer a lot of tensor shapes, deriving them both from dynamic dimension arguments (the first two lines of parameters to the function) and from static type-level hints.
```
tomsmeding, line 125:
```suggestion
An additional feature of this library is a type system for tensor shape arithmetic. The following code is part of a convolutional neural network definition, for which horde-ad computes the shape of the gradient from the shape of the input data and the initial parameters. The compiler is able to infer many tensor shapes, deriving them both from dynamic dimension arguments (the first two lines of parameters to the function) and from static type-level hints.
```

tomsmeding (commented):
I added some comments. Feel free to do with them what you wish.

Other suggestions that I couldn't place in-line because they were outside the diff area of Andrew's commit
```diff
-A quick inspection of the gradient code reveals that computations are not repeated, which is thanks to the sharing mechanism, as promised above.
+A quick inspection of the gradient code reveals that computations are not repeated, which is thanks to the `tlet` used above.
```

After the code snippet after this sentence:
> The value is the same as it would be for augmented `fooLet` evaluated by `cgrad` on the same input, as long as the incoming cotangent supplied for the interpretation consists of ones in all array cells, which is denoted by `srepl 1` in this case:

I think it would be good to state that this coincidence, where passing `srepl 1` as the initial cotangent allows removing the `kfromS . ssum0` from the objective function, is specifically because it was `ssum0`, and the reverse derivative of `ssum0` is replication. I'm not sure how to formulate this compactly.

README.md
Mikolaj: I think your remark about `atan2H` is sufficient (and in reality it's not different from `atan2` --- it just belongs to a class that contains solely `atan2H` and doesn't have a lot of silly superclasses). This is getting quite long already, so I'd keep it as you wrote it and maybe I'd even try to trim the README later on and move some of the contents to a hypothetical future tutorial.

README.md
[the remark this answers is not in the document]
Mikolaj: Toms says this should be doable and fits the user API fine (as opposed to the engine maybe), so I will do that. Thank you.

README.md
Mikolaj: Agreed. @tomsmeding I've made this compile, but I'm afraid it doesn't now look as well as we thought it would. In particular, we introduce `ADReady` and `RealFloat` of an array earlier than before. Which actually shows that even without AST the user has to touch the machinery only added due to AST, which is unfortunate, but maybe let's not scare the user without first showing the shiny features that justify the extra mental cost. The snippet I'm talking about is after the line, with a few lines of context. Shall I maybe merge and let you tweak locally afterwards?

-----------------------

```hs
fooMatrixValue :: Matrix2x2 Concrete Double
fooMatrixValue = foo threeSimpleMatrices
>>> fooMatrixValue
sfromListLinear [2,2] [4.242393641025528,4.242393641025528,4.242393641025528,4.242393641025528])
```

Instantiated to matrices, `foo` now returns a matrix, not a scalar &mdash; but a gradient can only be computed of a function that returns a scalar. Thus, let's define `sumFooMatrix` which sums the output of `foo`: (note that `ssum0` returns a zero-dimensional array; `kfromS` extracts the (single) scalar from that)
```hs
sumFooMatrix :: (ADReady f, RealFloat (Matrix2x2 f r), GoodScalar r)
             => (Matrix2x2 f r, Matrix2x2 f r, Matrix2x2 f r) -> f (TKScalar r)
sumFooMatrix = kfromS . ssum0 . foo
>>> sumFooMatrix threeSimpleMatrices
16.96957456410211
```

For this function we can compute the gradient as follows:
```hs
gradSumFooMatrix :: (Differentiable r, GoodScalar r)
                 => ThreeMatrices r -> ThreeMatrices r
gradSumFooMatrix = cgrad sumFooMatrix
```

This works as well as before:
```

File: README.md (hunk 2 above)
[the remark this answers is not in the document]
Comment (Mikolaj) [reply README.md:124]: Yes, the simplified source code of the gradient is in `artifact` and the interpreter is very straightforward, not doing any simplification, so the gradient is certainly only simplified once. If one of the tensors in the `threeOtherSimpleMatrices` tuple was ranked and if it's shape was different than the shape of the corresponding tensor in `threeSimpleMatrices`, then `vjpInterpretArtifact` would signal an error instead of redoing AD and then simplification (even though, by accident, it could, because `artifact` contains a slightly mangled source code of the objective function emitted as a natural side-effect of dual numbers AD).

**Comment by Mikolaj** [comment]
> I think it would be good to state that this coincidence, where passing `srepl 1` as the initial cotangent allows removing the `kfromS . ssum0` from the objective function, is specifically because it was `ssum0`, and the reverse derivative of `ssum0` is replication. I'm not sure how to formulate this compactly.

@tomsmeding: I'm considering removing this whole sentence instead, because my point was that if you amend your function with the obvious/cheap/lazy `kfromS . ssum0`, then this is interchangeable with the obvious/cheap/lazy `srepl 1`. But in the new text, the point is lost, because the function is not amended on the fly in the cheap way and instead a new serious function `sumFooMatrix` is introduced and the `kfromS . ssum0` looks like a very deliberate choice.

**Comment by Mikolaj** [comment]
I've incorporated all your changes and I only did minimal edits for now. The rendered readme can be viewed at https://github.com/Mikolaj/horde-ad/blob/9a9283c9aba8fa7e5669c8d7b7cf7f00a53dfb74/README.md

File: README.md (hunk 4 above)
Comment (tomsmeding) [reply README.md]: Do I understand correctly that you need this `ADReady` etc. here because we're going to instantiate it to something non-`Concrete` later -- namely `ADVal` for `cgrad`?

Comment (Mikolaj) [reply README.md]: Yes, in the next code snippet above, the one starting with `gradSumFooMatrix`, we instantiate it to `ADVal`, because we apply `cgrad` to it.

```hs
cgrad
  :: forall advals r.
     ( X advals ~ X (DValue advals), KnownSTK (X advals)
     , AdaptableTarget (ADVal Concrete) advals
     , AdaptableTarget Concrete (DValue advals) )
  => (advals -> ADVal Concrete (TKScalar r))
  -> DValue advals
  -> DValue advals  -- morally DValue (ADTensorKind advals)
```

README.md
```diff
@@ -61,13 +84,15 @@ fooLet (x, y, z) =
     atan2H z w + z * w  -- note that w still appears twice
 ```
 
+### Vector-Jacobian product (VJP)
+
 The most general symbolic gradient program for this function can be obtained using the `vjpArtifact` tool. We are using `fooLet` without `ssum0` this time, because the `vjp` family of tools by convention permits non-scalar codomains (but expects an incoming cotangent argument to compensate, visible in the code below as `dret`).
```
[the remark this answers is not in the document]
Mikolaj: I meant the convention in the wider world, just as jvp and forward derivatives usually take the extra argument, but I can see how this is misleading. Committed.

README.md
tomsmeding: I'm not sure how to rewrite this. I think it's quite explainable what's going on, but it may take a few sentences, and they may distract the reader at this point. Maybe, for now, just add a remark after the definition of `sumFooMatrix` like "We had to generalise the type of this function from `Concrete` to an arbitrary horde-ad target space `f` in order to differentiate it with `cgrad`. Under the hood, `cgrad` instantiates the function to `ADVal Concrete`: dual numbers of concrete tensors."

Mikolaj: Or maybe backtrack a little and avoid the new concepts by showing less code, as in the original version? Probably inlining `sumFooMatrix` would be enough to hide the worst offenders.

Mikolaj:
Thank you very much again. I'm merging.

